PR #10, opened 2021-03-06
Feature/chat

ahmedc3ca:
Once logged in you can press on the top right screen to start a new chat!

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,3 +1,4 @@
+
 plugins {
     id 'com.android.application'
     id 'kotlin-android'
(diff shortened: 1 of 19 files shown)


app/build.gradle
```diff
@@ -86,4 +93,7 @@ task jacocoTestReport(type: JacocoReport, dependsOn: ['testDebugUnitTest', 'crea
 
 connectedCheck {
     finalizedBy jacocoTestReport
-}
\ No newline at end of file
+}
+
+apply plugin: 'com.android.application'
```
MarcelTorne, line 98: Correct me if I am wrong but I think we can add these two lines on the top inside plugins in the following manner:
id 'com.android.application' (which can be removed because it is already present)
id 'com.google.gms.google-services'

app/src/main/java/com/github/marceltorne/meili/LoginActivity.kt
```diff
@@ -0,0 +1,53 @@
...
+            loginUser()
+        }
+
+        findViewById<TextView>(R.id.back_to_registration_text_view).setOnClickListener {
+            //Launch the login activity
+            finish()
+        }
+    }
+
+    private fun loginUser() {
```
MarcelTorne, line 30: I think it would be a good idea to call this function a little bit different, for example "simpleLoginUser" because we will also have googleLogin and then it could be confusing. What do you think?

MarcelTorne (commented):
I didn't see any tests, did I miss them? Otherwise, I don't think we will be able to merge since we setup everything so that we need at least 80% coverage to merge.

And thanks for the work seems like a lot of progress!

PR #12, opened 2021-03-08
rename package to com.github.epfl.meili

aman3014:


MarcelTorne (approved):

aman3014:
I think it is normal for the tests to fail since we are renaming the package. The tests will be updated by the renaming too.

Bwananah (approved):

PR #16, opened 2021-03-11
Implementation of Google Sign In 

MarcelTorne:


.idea/gradle.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ahmedc3ca, line 1: Do you think it's necessary to include the .idea/gradle.xml on the remote version?

ahmedc3ca (approved):
Everything Looks great, good job! Only remark is about the inclusion of the .idea/gradle.xml file on the remote

aman3014:
Looks good! Just don't forget to reformat the code using Android Studio (on Windows it's Ctrl+Alt+L) 😄

.idea/gradle.xml
MarcelTorne: No, indeed, I am going to remove it.

aman3014 (approved):

PR #18, opened 2021-03-11
Feature/posts

Bwananah:
Fixed the test, cleaned up some code and merged with main branch


app/src/main/java/com/github/epfl/meili/MainActivity.kt
```diff
@@ -4,16 +4,26 @@ import android.content.Intent
...
 
-    fun onClickLaunchSignInView(view: View){
+
+    fun onClickLaunchSignInView(view: View) {
         val intent = Intent(this, GoogleSignInActivity::class.java)
         startActivity(intent)
     }
+
+    /** Called when the user taps the forum button */
+    fun openForum(view: View) {
```
MarcelTorne, line 25: From my opinion it would be great to have 1 test for this button even if it's just temporary. On the MainActivityTest class there is already a Test made for the launchSignInView, you could reuse it.

app/src/androidTest/java/com/github/epfl/meili/NewPostTitleAndTextTest.kt
```diff
@@ -0,0 +1,193 @@
...
+@LargeTest
+@RunWith(AndroidJUnit4::class)
+class CanCreateNewPostTest {
+
+    @Rule
+    @JvmField
+    var mActivityTestRule = ActivityTestRule(MainActivity::class.java)
+
+    @Test
+    fun canCreateNewPostTest() {
```
MarcelTorne, line 41: Do you think it would be a good idea to split this test so that if it ever fails we can immediately identify what is failing from the test name? And maybe if it makes sense add some assertions every time you perform an action like clicking?

app/src/main/res/layout/activity_main.xml
```diff
@@ -13,8 +13,16 @@
         android:onClick="onClickLaunchSignInView"
         android:text="Launch Sign In View"
         app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintEnd_toEndOf="parent" />
+
+        <Button
+        android:id="@+id/main_forum_button"
```
MarcelTorne, line 19: When I render the app on my computer I see that there is a problem with the constraints of the `LaunchSignInView`, if it's also the case for you could you add some basic constraints to the Forum button and sides for example?

MarcelTorne (changes requested):
Really good job! Thanks for working on this! I wrote some comments

PR #19, opened 2021-03-12
Multiple activities for chatting interface

ahmedc3ca:


app/src/main/java/com/github/epfl/meili/messages/NewMessageActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+
+
+        fetchUsers()
+    }
+
+    companion object {
+        val USER_KEY = "USERK_KEY"
+    }
+
+    private fun fetchUsers() {
```
MarcelTorne, line 33: I think it would be good for modularisation to have this in an auxiliary ViewModel class. But since this is already a huge PR maybe you can add in the Backlog a task that asks for refactoring of this code and leave it for later.

MarcelTorne: I am saying this because I believe that ideally the Activities should only do UI related actions and not the actions that correspond to the model as fetching users from database.

MarcelTorne (commented):
That's a lot of work done, thanks! I added two comments, one is regarding the MVVC model that I think we should implement but I propose that you leave it for another Task that we can add in the Backlog

app/src/main/java/com/github/epfl/meili/messages/NewMessageActivity.kt
ahmedc3ca: Totally agree! Refactoring the code to comply with the MVCC model will be my top priority for the next sprint.

app/src/main/java/com/github/epfl/meili/messages/ChatLogActivity.kt
```diff
@@ -0,0 +1,119 @@
...
+                if (chatMessage != null) {
+                    if (chatMessage.fromId == FirebaseAuth.getInstance().uid) {
+                        adapter.add(ChatItem(chatMessage.text, true))
+                    } else if (chatMessage.toId == toId) {
+                        adapter.add(ChatItem(chatMessage.text, false))
+                    }
+                }
+            }
+
+            override fun onChildChanged(snapshot: DataSnapshot, previousChildName: String?) {
```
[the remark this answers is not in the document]
ahmedc3ca: Unfortunately yes since the ChildEventListener is an abstract class I have to override all the methods. I tried finding workarounds but this seems to be the go-to solution on stack-overflow.

aman3014 (approved):

app/src/main/AndroidManifest.xml
```diff
@@ -17,6 +20,19 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity android:name=".ChatLogActivity">
```
MarcelTorne, line 23: Can you change this to MainActivity and add ChatLogActivity below with the rest of activities please? (discussed internally)

MarcelTorne (approved):
Really good job! Thank you!

PR #47, opened 2021-03-14
Bug/sign in test

MarcelTorne:


aman3014 (approved):
Thanks!

app/src/androidTest/java/com/github/epfl/meili/home/GoogleSignInActivityTest.kt
```diff
@@ -24,53 +21,68 @@ import org.junit.runner.RunWith
...
+        onView(withId(R.id.signInButton))
+                .check(matches(isClickable())).perform(click())
+
+        val mockIntent = mockService.signInIntent()
+
+        Intents.intended(IntentMatchers.filterEquals(mockIntent))
+
+        Intents.release()
+    }*/
+
```
tberkane, line 80: Is it normal that this test is commented out?

app/src/androidTest/java/com/github/epfl/meili/home/MockAuthenticationService.kt
```diff
@@ -0,0 +1,34 @@
+package com.github.epfl.meili.home
+
+import android.app.Activity
+import android.content.Intent
+
+class MockAuthenticationService : AuthenticationService {
+    var MOCK_NAME = "Meili User"
```
tberkane, line 7: I think this should be a private const val, and be in a companion object.

app/src/main/java/com/github/epfl/meili/home/FirebaseAuthenticationService.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.GoogleAuthProvider
+import com.google.firebase.auth.ktx.auth
+import com.google.firebase.ktx.Firebase
+
+class FirebaseAuthenticationService : AuthenticationService {
+    private lateinit var auth: FirebaseAuth
+    private lateinit var googleSignInClient: GoogleSignInClient
+    private val TAG = "GoogleActivity"
+    private val RC_SIGN_IN = 9001
```
tberkane, line 21: const val + companion object ?

tberkane (changes requested):
Good job as always! Keep up the good work! Just a few nitpicks.

app/src/androidTest/java/com/github/epfl/meili/home/GoogleSignInActivityTest.kt
```diff
@@ -101,29 +113,13 @@ class GoogleSignInActivityTest {
...
-
-        testRule!!.scenario.onActivity { activity ->
-            AuthenticationService.firebaseAuthWithGoogle(activity!!, fake_id)
-
-            assert(!AuthenticationService.isLoggedIn.value!!)
-            assert(AuthenticationService.name == null)
-            assert(AuthenticationService.email == null)
-        }
+        intending(IntentMatchers.filterEquals(mockIntent)).respondWith(result)
+        Intents.release()
```
aman3014, line 123: I suggest putting Intents.init() in a separate function with the @Before tag and Intents.release() in one with the @After tag since intents are being initialized and released for multiple tests.

aman3014 (changes requested):

app/src/androidTest/java/com/github/epfl/meili/home/GoogleSignInActivityTest.kt
MarcelTorne: This test should be removed. Thanks for pointing it out!

app/src/androidTest/java/com/github/epfl/meili/home/MockAuthenticationService.kt
MarcelTorne: I will change this! Do you know what's the policy for using companion objects? Like, when should we be using them?

MarcelTorne: Actually, in this case, I was modifying the values in the test. So they weren't constants, I removed the capital letters which were indeed misleading

tberkane (approved):
Thanks for fixing the small nitpicks!

aman3014 (approved):

Bwananah (approved):

PR #48, opened 2021-03-14
Implement basic google maps view which shows location and POIs

### LINKED ISSUE #2: As a traveller, I want to see a map with my position so I can know my current position

### LINKED ISSUE #5: As a traveller, I want to see points of interest around me so that I can know what interesting things there are to see nearby

### LINKED ISSUE #49: Automatically center map on user's location

aman3014:
Contributions by @tberkane and @aman3014 merged into one commit due to merge issues in the original branch.
Closes #2
Closes #5 
Closes #49


app/src/main/java/com/github/epfl/meili/MapActivity.kt
```diff
@@ -0,0 +1,145 @@
...
+        } else {
+            map.isMyLocationEnabled = false
+            map.uiSettings?.isMyLocationButtonEnabled = false
+            location = null
+            Toast.makeText(this, "Cannot show location without permission", Toast.LENGTH_LONG)
+        }
+    }
+
+    @SuppressLint("MissingPermission")
+    private fun getDeviceLocation() {
```
MarcelTorne, line 116: Since I believe that we will be using the device location in a more parts of the project would it be possible that you created a Service that provides the location? What do you think?

app/src/androidTest/java/com/github/epfl/meili/MapActivityTest.kt
```diff
@@ -0,0 +1,92 @@
...
+//    @Test
+//    fun a_shouldDisplayPermissionRequestDialogAtStartup() {
+//        val device = UiDevice.getInstance(getInstrumentation());
+//        Thread.sleep(2000)
+//        assertViewWithTextIsVisible(device, "ALLOW")
+//        assertViewWithTextIsVisible(device, "DENY")
+//
+//        // cleanup for the next test
+//        reactToPermission(device, "DENY")
+//    }
```
tberkane, line 54: This commented out code should be removed

app/src/androidTest/java/com/github/epfl/meili/MapActivityTest.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    fun b_locationButtonClickableAfterPermissionGrant() {
+        PermissionGranter.allowPermissionsIfNeeded("android.permissions.ACCESS_FINE_LOCATION")
+//        Thread.sleep(2000)
```
tberkane, line 59: Here as well

MarcelTorne (changes requested):
I added several comments. It is a shame regarding the testing coverage but I believe that you have already tried to find many ways on how to improve it so I guess it's okay. Thank you for your work!

app/src/androidTest/java/com/github/epfl/meili/MapActivityTest.kt
```diff
@@ -0,0 +1,92 @@
...
+        val imageView = onView(
+            allOf(
+                withContentDescription("My Location"),
+                childAtPosition(
+                    childAtPosition(
+                        withClassName(`is`("android.widget.FrameLayout")),
+                        2
```
tberkane, line 66: Maybe use named constants here

app/src/main/java/com/github/epfl/meili/MapActivity.kt
```diff
@@ -0,0 +1,145 @@
...
+            Manifest.permission.ACCESS_FINE_LOCATION
+        ) == PackageManager.PERMISSION_GRANTED
+    }
+
+    override fun onRequestPermissionsResult(
+        requestCode: Int,
+        permissions: Array<String>,
+        grantResults: IntArray
+    ) {
+//        // Check whether permission was granted
```
tberkane, line 70: Commented out code to remove

tberkane (changes requested):
Good job! Just a few nitpicks.

Bwananah (approved):
Looks OK to me. Should mybe put into new packages but that can be done later

app/src/androidTest/java/com/github/epfl/meili/MainActivityTest.kt
```diff
@@ -1,42 +1,73 @@
...
+//    @Test
+//    fun testNavigation() {
+//        PermissionGranter.allowPermissionsIfNeeded("android.permissions.ACCESS_FINE_LOCATION")
+//        onView(withId(R.id.launchSignInView)).perform(click())
+//        onView(withId(R.id.signInButton)).check(matches(isDisplayed()))
+//        pressBack()
+//        onView(withId(R.id.launchMap)).perform(click())
+//        onView(withId(R.id.map)).check(matches(isDisplayed()))
+//        pressBack()
+//        onView(withId(R.id.launchSignInView)).check(matches(isClickable())).perform(click())
```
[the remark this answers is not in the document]
aman3014: It is commented on purpose. This is because the test works on a device (where no permissions are given by default) but doesn't work with gradlew as the permission dialog box doesn't appear. I suppose this is because the permission is given  by default and I haven't found a way to revoke permissions at the start of each test.

app/src/androidTest/java/com/github/epfl/meili/MapActivityTest.kt
aman3014: The test was generated using Espresso Test Recorder. Do you think it still makes sense to name these constants?

app/src/androidTest/java/com/github/epfl/meili/MapActivityTest.kt
aman3014: I wanted to leave this test commented since it does work with a device, just not with gradlew. What do you think?

aman3014: I can add a comment in the code saying the same.

app/src/main/java/com/github/epfl/meili/MapActivity.kt
aman3014: Yes, I can try to do that.

app/src/androidTest/java/com/github/epfl/meili/MainActivityTest.kt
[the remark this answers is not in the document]
aman3014: I've added this comment to the code too.

app/src/main/java/com/github/epfl/meili/MapActivity.kt
aman3014: In fact, the MapActivity is already using a LocationService provided by android. I tried abstracting away the permission handling in another class. However, I realized that it has to be done in whichever activity is using the location because it needs to be updated when the permission is granted.

app/src/androidTest/java/com/github/epfl/meili/MainActivityTest.kt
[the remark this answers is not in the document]
MarcelTorne: Okay Thanks!

app/src/main/java/com/github/epfl/meili/MapActivity.kt
MarcelTorne: Couldn't we just ask for the permission at the very beginning after they login or when they start the app? Cause in the Sweng mini-project we managed to make a new service for the Location if I recall correctly, no?

aman3014: They were not using the fusedLocationProviderClient in the mini project. Here it does most of the job for us. As for the permissions, yes we could. But if we are only doing it once, I don't think we should factor it out. In the mini project last semester, the permission handling was done in the weather activity itself.

aman3014:
> Looks OK to me. Should mybe put into new packages but that can be done later

Done

app/src/main/java/com/github/epfl/meili/MapActivity.kt
MarcelTorne: I think we will still have to provide a LocationService but let's leave this as next steps. Thanks!

MarcelTorne (approved):

tberkane (approved):
gj

PR #50, opened 2021-03-16
Implement MessageDatabase and Firebase corresponding object

MarcelTorne:
This is the first step towards resolving issue #46

### DIFF
--- a/app/src/androidTest/java/com/github/epfl/meili/messages/FirebaseMessageDatabaseAdapterTest.kt
+++ b/app/src/androidTest/java/com/github/epfl/meili/messages/FirebaseMessageDatabaseAdapterTest.kt
@@ -3,41 +3,40 @@ package com.github.epfl.meili.messages
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.internal.runner.junit4.statement.UiThreadStatement
 import com.github.epfl.meili.models.ChatMessage
-import junit.framework.Assert
 import junit.framework.Assert.assertEquals
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
-import java.lang.IllegalArgumentException
 import java.util.*
-import kotlin.collections.ArrayList
 
 @RunWith(AndroidJUnit4::class)
 class FirebaseMessageDatabaseAdapterTest {
     private val MOCK_PATH = "POI/mock-poi"
     private lateinit var db: FirebaseMessageDatabaseAdapter
-    private val fake_message = ChatMessage("fake_id", "fake_text", "fake_from_id", "fake_to_id", 12345)
+    private val fake_message =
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/epfl/meili/messages/FirebaseMessageDatabaseAdapter.kt
```diff
@@ -0,0 +1,55 @@
...
+    override fun onChildAdded(snapshot: DataSnapshot, previousChildName: String?) {
+        val chatMessage = snapshot.getValue(ChatMessage::class.java)
+
+        if (chatMessage != null) {
+            messages.add(chatMessage)
+            this.notifyObservers()
+        }
+    }
+
+    override fun onChildChanged(snapshot: DataSnapshot, previousChildName: String?) {
```
Bwananah, line 44: Is it necessary to override without replacing it with anything ?

app/src/androidTest/java/com/github/epfl/meili/messages/FirebaseMessageDatabaseAdapterTest.kt
```diff
@@ -0,0 +1,53 @@
...
+class FirebaseMessageDatabaseAdapterTest {
+    private val MOCK_PATH = "POI/mock-poi"
+    private lateinit var db: FirebaseMessageDatabaseAdapter
+    private val fake_message =
+        ChatMessage("fake_id", "fake_text", "fake_from_id", "fake_to_id", 12345)
+
+    @Before
+    fun initializeDatabase() {
+        UiThreadStatement.runOnUiThread {
+            db = FirebaseMessageDatabaseAdapter(MOCK_PATH)
```
Bwananah, line 22: What is MOCK_PATH ? Is it used to create a mock database ?

Bwananah (commented):
Very good job. You are a machine ! I would prefer if you added more comments to explain your code as I am not as familiar with android or Firebase. But I should get better at reading code as the semester continues.

I am waiting until you see my questions before I approve, but it should be fine

app/src/main/java/com/github/epfl/meili/messages/FirebaseMessageDatabaseAdapter.kt
MarcelTorne: Unfortunately yes it is

app/src/androidTest/java/com/github/epfl/meili/messages/FirebaseMessageDatabaseAdapterTest.kt
MarcelTorne: I will write it as comment, path is the path of the Firebase database, so yes it is for a mock database

Bwananah (approved):
Thanks for the changes, good job !

ahmedc3ca (approved):
A great addition to the code. I will need to rebase this to my branch to use the firebaseMessageAdapter.

PR #55, opened 2021-03-18
Feature/forum

Bwananah:
Blank pull request, I need help on how to make my tests start


app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
```diff
@@ -0,0 +1,60 @@
...
+    override suspend fun getPosts(): List<Post> {
+        return posts
+    }
+
+    override fun addPost(author: String, title: String, text: String) {
+        posts.add(Post((idCounter++).toString(),author, title, text))
+        notifyObservers()
+    }
+
+    override fun addObserver(o: Observer?) {
```
MarcelTorne, line 45: Instead of implementing the observer methods here, don't you think it would be better to have it in PostService? So that the children can inherit it?

app/src/androidTest/java/com/github/epfl/meili/models/ChatMessageUnitTest.kt
```diff
@@ -1,41 +0,0 @@
-package com.github.epfl.meili.models
```
MarcelTorne, line 1: Why is this file deleted? It should be preserved right?

app/src/androidTest/java/com/github/epfl/meili/models/UserUnitTest.kt
```diff
@@ -1,70 +0,0 @@
-package com.github.epfl.meili.models
```
MarcelTorne, line 1: Same here, shouldn't we keep it?

app/build.gradle
```diff
@@ -56,24 +55,19 @@ dependencies {
     implementation 'com.google.firebase:firebase-analytics-ktx'
     implementation 'com.google.firebase:firebase-auth-ktx'
     implementation 'com.google.firebase:firebase-database-ktx'
+    implementation 'com.google.firebase:firebase-firestore-ktx'
     implementation "com.xwray:groupie:2.9.0"
+    implementation 'com.android.support.test.espresso:espresso-idling-resource:3.0.2'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-play-services:1.4.1'
-    implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.3.0"
```
MarcelTorne, line 61: I would be careful in deleting so many dependencies, are you sure these are not needed somewhere else in the project?

MarcelTorne (changes requested):
Awesome job! Thank you very much! This has been a hard one but we are super close to the end of it! So keep up the good work!

(I wrote a couple of comments and I hope they'll help)

app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
Bwananah: This is an Observable method. The changes to the Observable methods only apply to the Mock Service, so I think it is better there. PostService already inherits from Observable(), so FirebasePostService just uses the Observable() default methods.

app/src/androidTest/java/com/github/epfl/meili/models/ChatMessageUnitTest.kt
Bwananah: I have no idea how this happened. I know I deleted tests to see if mine passed, but I thought I had restored them. I will put them back in.

Bwananah:
I  removed the Poi test, because it failed even though its not supposed to do anything. I updated the files from main and it still fails the empty test, which is weird.

app/build.gradle
Bwananah: I updated it.

Bwananah: My b, I must have not rebased correctly, I had a lot of git problems.

app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
MarcelTorne: Sorry, maybe I am not understanding correctly, but I don't really understand why you would use default methods from Observable in FirebasePostService but not in the mock service. Could you clarify a little bit more, please? Thanks!

app/src/test/java/com/github/epfl/meili/helpers/DateAuxiliaryTest.kt
```diff
@@ -0,0 +1,28 @@
+package com.github.epfl.meili.helpers
```
MarcelTorne, line 1: This is a little bit weird, it seems like the changes from previous PRs are being committed again. Do you have any idea of what might be going wrong and how to fix it?

MarcelTorne (changes requested):
Thanks again! It seems like there are some problems with the branches since files from other PRs are being added here too. Maybe rebasing would fix it?

app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
Bwananah: I'm not quite sure, Ahmed did this part. But I tried what you said and the mock service did not end up working.

app/src/test/java/com/github/epfl/meili/helpers/DateAuxiliaryTest.kt
Bwananah: I can't rebase anymore, there are way too many commits. I tried for 2 hours to find a solution to it and mine was to just copy the changes from main to make sure there aren't going to be any conflicts.

Bwananah: Like if I try to rebase, there are conflicts on every single commit, I don't know why

Bwananah:
I don't know why it fails the Log in test, I updated it manually from main.

app/src/test/java/com/github/epfl/meili/helpers/DateAuxiliaryTest.kt
Bwananah: I have way too many commits and when I rebase there are conflicts on every single one of them. I tried squashing and other git stuff without success. my solution was to just make sure there aren't going to be any conflicts when merging.

PR #56, opened 2021-03-18
Add button to main activity to access poi activity.

tberkane:


app/src/androidTest/java/com/github/epfl/meili/poi/PoiActivityTest.kt
```diff
@@ -0,0 +1,24 @@
...
+@LargeTest
+@RunWith(AndroidJUnit4::class)
+class PoiActivityTest {
+
+    @get: Rule
+    var testRule = ActivityScenarioRule(PoiActivity::class.java)
+
+    @Test
+    fun poiActivityTest() {
+
```
ahmedc3ca, line 20: Adding a  //todo here and explaining what needs to be done can be beneficial since someone can recognize he did something similar before and help. Also, Android Studio has a tool that shows you all the marked TODOs and it makes it easier to manage the work.

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
```diff
@@ -81,15 +81,23 @@ class MapActivity : AppCompatActivity(), OnMapReadyCallback {
         } else {
             getLocationPermission()
         }
+
+        //map.setOnPoiClickListener(this)
     }
 
+    /*override fun onPoiClick(poi: PointOfInterest) {
```
ahmedc3ca, line 88: Should I uncomment this method when I rebase to make clicking the poi go to the poi chat log?

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
```diff
@@ -24,9 +24,9 @@ import com.google.android.libraries.places.api.net.PlacesClient
 
 class MapActivity : AppCompatActivity(), OnMapReadyCallback {
     companion object {
-        private val TAG = MapActivity::class.java.name
         private const val DEFAULT_ZOOM = 15
         private const val REQUEST_CODE: Int = 1
+        const val POI_KEY = "POI_KEY"
```
aman3014, line 29: Could you please explain what the POI_KEY is supposed to represent?

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
aman3014, line 88: Why is this code commented out?

ahmedc3ca (approved):
Good Job! The POI on the map look great. Also, this PR is important for my branch since the app will appear more functional when the chat log appears when you click on a POI.

app/src/main/res/layout/activity_main.xml
```diff
@@ -43,4 +43,16 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/launchSignInView" />
 
+    <Button
+        android:id="@+id/launchPoiView"
```
aman3014, line 47: I suggest removing this button and integrating the poi view with the map. What do you think?

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
aman3014, line 88: cf comment on activity main.

aman3014: cf comment on activity_main.xml

aman3014 (changes requested):

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
tberkane: Yes, for now this is commented because I haven't figured out yet how to test it and I want to be able to merge

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
tberkane: It's used as a key when passing the Intent (which contains the poi as a value) to the PoiActivity which is launched when a poi on the map is clicked.

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
tberkane: Because I can't figure out how to test it and I want to merge

app/src/main/res/layout/activity_main.xml
tberkane: Yes but then onPoiClick has to be tested I think, which is tricky

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
MarcelTorne: One question, in the ChatGroup section we expect to get a uniquely identifier for the POI clicked. Do you think this could the best place to give this uid in the intent? If yes, are you planning to do it as a follow up? If no, where do you think it would be better to do this?

MarcelTorne (commented):
Really cool and really good job! Thank you! I asked a question below regarding the POI_KEY.

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
ahmedc3ca: I think it's ok to send the entire POI for now since I also use the POI name to show at the top of the chat log activity. Also, we might want to be able to show other features of the POI on the chat interface later.

tberkane: @MarcelTorne Yes I think this is the right place, and actually Google Map's PointOfInterest class already contains such a unique id as a field. Ad @ahmedc3ca says, I am passing the entire POI here so that we can also access its name, etc. as needed.

MarcelTorne: That sounds really good, thank you!

aman3014 (approved):

PR #66, opened 2021-03-22
Bug: Remove Register and Login Button Tests

MarcelTorne:
The tests for these two buttons are failing when ran on Cirrus because a Google popup asking wether you want to remember the password appears.


ahmedc3ca (approved):
It's good to remove the tests that don't pass the cirrus ci for now but we would have to make up in code coverage later.

aman3014 (approved):

MarcelTorne:
@aman3014 It fixes part of the problems, now it doesn't fail due to not finding the views (which was due to the google pop up appearing) but when it fails it is because it didn't respond within the 10sec limit (at least this is what I've seen). And I don't think we can do anything to fix this problem, it is probably due to CirrusCI, moreover, we also had this errors
 at the very beginning.

aman3014:
Okay, I see. Thanks for the explanation.

aman3014 (approved):

MarcelTorne:
Maybe we could try to add:
```
MyActivity activityUnderTest = activityTestRule.getActivity();
activityUnderTest.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
```
But I don't know how to do it for all of the tests without having to copy paste it one by one.
(source: https://stackoverflow.com/questions/39457305/android-testing-waited-for-the-root-of-the-view-hierarchy-to-have-window-focus/40015359)

aman3014:
Do you think the @BeforeEach annotation could work?

MarcelTorne:
Yes, but we'll have to do it in all of the test files, no?
Anyways, let me try to add it in Register and Login and we'll see if the error continues to appear.
Thanks for pointing it out!

PR #70, opened 2021-03-24
data classes should be covered by tests now

ahmedc3ca:


MarcelTorne:
That's really great! Thanks Ahmed!
I propose that everyone contributes to this PR by adding test coverage of one data class and then we merge.
What do you think @ahmedc3ca?

PR #88, opened 2021-03-28
POI Service

MarcelTorne:
The POIService returns all of the points of interest that are close to the location passed as parameter. The information is taken using Overpass API from OpenStreetMap.


aman3014 (changes requested):
Everything looks good, just I think we should make as many fields private as possible. I put a comment on a (potential) example field which could be made private.

aman3014 (approved):
Looks great, thanks!

tberkane (approved):
Great job!

PR #89, opened 2021-03-29
Merge/forum

Bwananah:
Features are from feature/forum. I had a lot of trouble with git, and rebasing. This is the cleanest way I found to make a pull request that wasn't 70 commits long and that included changes made in main.

### DIFF
--- a/app/src/androidTest/java/com/github/epfl/meili/forum/PostActivityAndroidTest.kt
+++ b/app/src/androidTest/java/com/github/epfl/meili/forum/PostActivityAndroidTest.kt
@@ -1,0 +1,202 @@
+package com.github.epfl.meili.forum
+
+
+import android.view.View
+import android.view.ViewGroup
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.Espresso.pressBack
+import androidx.test.espresso.action.ViewActions.*
+import androidx.test.espresso.intent.Intents
+import androidx.test.espresso.matcher.ViewMatchers.*
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement
+import com.github.epfl.meili.MainActivity
+import com.github.epfl.meili.R
+import com.github.epfl.meili.forum.Post.Companion.toPost
+import com.github.epfl.meili.home.Auth
+import com.github.epfl.meili.home.MockAuthenticationService
+import com.google.android.gms.tasks.Task
+import com.google.firebase.firestore.*
+import org.hamcrest.Description
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.TypeSafeMatcher
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito
+
+
+@RunWith(AndroidJUnit4::class)
+class PostActivityAndroidTest {
+
+    private val TEST_TEXT = "test text"
+    private val TEST_TITLE = "test title"
+    private val TEST_USERNAME = "test_username"
+    private val TEST_EMAIL = "test@meili.com"
+    private val postList = ArrayList<Post>()
+    private val mockPost = Post("test id", TEST_USERNAME, TEST_TITLE, TEST_TEXT)
+    private val mockList = ArrayList<DocumentSnapshot>()
+
+    private lateinit var mockFirestore: FirebaseFirestore
+    private lateinit var mockDocumentSnapshot: DocumentSnapshot
+    private lateinit var mockCollectionReference: CollectionReference
+    private lateinit var mockDocumentReference: DocumentReference
+    private lateinit var mockQuerySnapshot: QuerySnapshot
+
+
+    @get:Rule
+    var testRule: ActivityScenarioRule<MainActivity> =
+        ActivityScenarioRule(MainActivity::class.java)
+
+    @Before
+    fun initializeMockDatabase() {
+        postList.add(mockPost)
+
+        mockFirestore = Mockito.mock(FirebaseFirestore::class.java)
+        mockDocumentSnapshot = Mockito.mock(DocumentSnapshot::class.java)
+        mockCollectionReference = Mockito.mock(CollectionReference::class.java)
+        mockDocumentReference = Mockito.mock(DocumentReference::class.java)
+        mockQuerySnapshot = Mockito.mock(QuerySnapshot::class.java)
+        val mockTask = Mockito.mock(Task::class.java)
+
+        Mockito.`when`(mockFirestore.collection("posts")).thenReturn(mockCollectionReference)
+        Mockito.`when`(mockCollectionReference.document(ArgumentMatchers.any())).thenReturn(mockDocumentReference)
+        Mockito.`when`(mockDocumentReference.get()).thenAnswer { mockDocumentSnapshot }
+        Mockito.`when`(mockCollectionReference.get()).thenAnswer { mockQuerySnapshot }
+        Mockito.`when`(mockCollectionReference.add(ArgumentMatchers.any())).thenAnswer{
+            mockList.add(mockDocumentSnapshot)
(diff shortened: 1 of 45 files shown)


app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    override fun notifyObservers() {
+        super.notifyObservers()
+
+        for(observer in observers){
+            observer.update(this, posts)
+        }
+    }
+}
```
Bwananah, line 60: Conversation from previous PR :  

@MarcelTorne :
Instead of implementing the observer methods here, don't you think it would be better to have it in PostService? So that the children can inherit it?

@Bwananah :
This is an Observable method. The changes to the Observable methods only apply to the Mock Service, so I think it is better there. PostService already inherits from Observable(), so FirebasePostService just uses the Observable() default methods.

@MarcelTorne :
Sorry, maybe I am not understanding correctly, but I don't really understand why you would use default methods from Observable in FirebasePostService but not in the mock service. Could you clarify a little bit more, please? Thanks!

@Bwananah :
I'm not quite sure, @ahmedc3ca did this part. But I tried what you said and the mock service did not end up working.

ahmedc3ca: For context: when I made the MockPostService, I created the PostService interface along with it. The FirebasePostService was already there. I didn't want PostService to implement the observer methods because then I would have to add the observables list to it. This means that FirebasePostService will also use all of that, but it didn't seem to me that Ewan was using them. So from my point of view, the interface PostService didn't necessarily use the observer's methods. If you think it is cleaner to have the interface implement the observer methods tho, it is easily arrangeable (just copy/paste from MockPostService).

app/src/main/java/com/github/epfl/meili/forum/NewPostViewModel.kt
```diff
@@ -0,0 +1,32 @@
+package com.github.epfl.meili.forum
```
ahmedc3ca, line 1: What do you think about having the PostViewModel, ForumViewModel & NewPostViewModel merged into one ViewModel since they are using the same service? But again, they are used for different activities. I am just curious to know your opinion about this.

ahmedc3ca (commented):
The tests work, the coverage is good and the functionality is great! Apart from the few aforementioned remarks, this should be a great addition to the main.

app/src/main/java/com/github/epfl/meili/forum/NewPostViewModel.kt
Bwananah: My guess is that they will vary a lot once we add more functionality to the forum, such as comments, upvotes, sorting methods and all that, so I think its better to keep them separate.

app/src/androidTest/java/com/github/epfl/meili/forum/PostActivityAndroidTest.kt
```diff
@@ -0,0 +1,202 @@
...
+                        withId(android.R.id.content),
+                        0
+                    ),
+                    2
+                ),
+                isDisplayed()
+            )
+        )
+        materialButton5.perform(click())
+    }
```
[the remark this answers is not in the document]
Bwananah: Yeah makes sense.
How do I make sure the suggestions do the same thing ? Because some times I have to make sure it's in the right position such as when I click on a post.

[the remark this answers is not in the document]
ahmedc3ca: Do you mean the espresso suggestions? My best guess is that you have to do them manually every time. Espresso might have more options that I don't know of tho.

app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
MarcelTorne: Then maybe the best would be to remove the Observable interface from PostService? Because if you say that FirebasePostService is not using it maybe we don't need it. Otherwise, if you see that you do need it then I think the best would be to implement the methods in PostService

MarcelTorne (changes requested):
There seems to be a small bug, when creating a new forum post, several post appear repeatedly.
There was tour Eiffel and forum is working post, but then when I add testing forum, the others appear repeated. Could you have a look at that, please?
Captura de Pantalla 2021-03-30 a les 14 21 51

Appart from that, awesome work, thanks for it!

app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
Bwananah: FirebasePostService is using the Observable interface. It uses notifyObservers() and setChanged()

MarcelTorne:
> There seems to be a small bug, when creating a new forum post, several post appear repeatedly.
> There was tour Eiffel and forum is working post, but then when I add testing forum, the others appear repeated. Could you have a look at that, please?
> Captura de Pantalla 2021-03-30 a les 14 21 51
> 
> Appart from that, awesome work, thanks for it!

Maybe having a look at `ChatLogActivity`class, `listenForMessages`function can help you fix the bug.

app/src/main/java/com/github/epfl/meili/forum/FirebasePostService.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    init {
+        // Listen to changes in post collection and notify observers when it has changed
+        db.collection("posts").addSnapshotListener { snapshot, e ->
+            // Handle errors
+            if (e != null) {
+                Log.w(TAG, "Listen failed.", e)
+                return@addSnapshotListener
+            }
+
```
MarcelTorne, line 29: In this line transform snapshot to list of ForumPosts and save it to service variable called posts: ArrayList<ForumPost>

app/src/main/java/com/github/epfl/meili/forum/FirebasePostService.kt
```diff
@@ -0,0 +1,79 @@
...
+                .toPost() // Convert to Post instance
+        } catch (e: Exception) {
+            Log.e(TAG, "Error getting Post from database")
+            null // Return null if error occurs
+        }
+    }
+
+    /** Get multiple posts from Database */
+    override suspend fun getPosts(): List<Post> {
+        return try {
```
MarcelTorne, line 55: Remove everything just return posts variable

app/src/main/java/com/github/epfl/meili/forum/MockPostService.kt
Bwananah: I have moved the Observable changes into PostService

app/src/androidTest/java/com/github/epfl/meili/forum/PostActivityAndroidTest.kt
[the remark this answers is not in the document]
Bwananah: It passes the automatic tests, so I think it's better to leave it like that for now

app/src/main/java/com/github/epfl/meili/forum/FirebasePostService.kt
aman3014: @Bwananah could you please explain the purpose of this listener? If I understood correctly you are using "get" on the collection reference whenever you want to get documents, so the listener isn't doing anything, right?

If you do what @MarcelTorne said above, you could actually get rid of the asynchronous "get" methods. That is how I implemented FirestoreReviewService in #53

aman3014: @Bwananah could you please explain the purpose of this listener? If I understood correctly you are using "get" on the collection reference whenever you want to get documents, so the listener isn't doing anything, right?

If you do what @MarcelTorne said above, you could actually get rid of the asynchronous "get" methods. That is how I implemented FirestoreReviewService in #53

aman3014 (commented):
cf comment in FirebasePostService

app/src/main/java/com/github/epfl/meili/forum/FirebasePostService.kt
Bwananah: I changed that yesterday thanks to Marcel. See the code below.

Bwananah:
Rebase changed too much stuff, making a new PR to merge

PR #90, opened 2021-03-29
Feature/poi info

tberkane:
Adds a PoiActivity, which can be launched by a button in the main activity.
Launching by clicking on a poi on the map will come in a later pr.


app/src/main/java/com/github/epfl/meili/poi/PoiInfoFragment.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private fun getOnSuccessListener(
+        placeId: String,
+        view: View,
+        placesClient: PlacesClient
+    ): (FetchPlaceResponse) -> Unit =
+        { response: FetchPlaceResponse ->
+            val place = response.place
+            Log.i(TAG, placeId)
```
aman3014, line 66: is this leftover from debugging

aman3014 (approved):

app/src/main/java/com/github/epfl/meili/poi/PoiActivity.kt
```diff
@@ -3,52 +3,54 @@ package com.github.epfl.meili.poi
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
-import androidx.fragment.app.FragmentActivity
 import androidx.viewpager2.adapter.FragmentStateAdapter
 import androidx.viewpager2.widget.ViewPager2
 import com.github.epfl.meili.R
 import com.github.epfl.meili.map.MapActivity
+import com.google.android.gms.maps.model.PointOfInterest
```
ahmedc3ca, line 10: I'm still crashing when clicking from the map view because of this discontinuity of POI versions I think.

ahmedc3ca (commented):
Looks very nice! I just can't demo from the map view. Also, for me, the main page is a bit wonky (the buttons are stacked on top of each other).

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
```diff
@@ -99,14 +99,11 @@ class MapActivity : AppCompatActivity(), OnMapReadyCallback {
 
         // Add on click listener
         clusterManager.setOnClusterItemClickListener {
-            val intent = Intent(this, PoiActivity::class.java)
```
MarcelTorne, line 102: Why did you remove this part? This was supposed to launch the PoiInfo activity when you click on a Poi

MarcelTorne (commented):
Seems really good to me! Thanks for the work! Left one small comment.

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
tberkane: I removed it because the query made in the poi info activity was not working (it didn't return anything). So I think that something doesn't work in the PointOfInterest which is passed in the intent by the MapActivity.

app/src/main/java/com/github/epfl/meili/poi/PoiActivity.kt
tberkane: It should be fixed now (I removed clicking on pois for now).

app/build.gradle
```diff
@@ -69,10 +70,11 @@ dependencies {
     implementation "androidx.activity:activity-ktx:1.2.2"
     implementation "androidx.fragment:fragment-ktx:1.3.2"
     implementation "androidx.viewpager2:viewpager2:1.0.0"
-
+    implementation 'com.google.android.material:material:1.3.0'
+    implementation "androidx.viewpager2:viewpager2:1.0.0"
```
aman3014, line 74: These are duplicate dependencies

app/src/androidTest/java/com/github/epfl/meili/poi/PoiActivityTest.kt
```diff
@@ -1,24 +1,87 @@
...
+        `when`(mockPlacesClient.fetchPhoto(any())).thenReturn(tcs2.task)
+
+        `when`(mockPlaces.getPlacesClient(any(), any())).thenReturn(mockPlacesClient)
+
+        PoiInfoFragment.placesClientService = { mockPlaces }
     }
 
+    @get:Rule
+    val mActivityTestRule: ActivityTestRule<PoiActivity> =
+        object : ActivityTestRule<PoiActivity>(PoiActivity::class.java) {
```
aman3014, line 71: Please use ActivityScenarioRule, ActivityTestRule is deprecated

aman3014 (changes requested):
Thanks for the work! Couple of minor changes requested

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
MarcelTorne: I see that's probably because the uid of the poi is from the overpass api and not from Google Places API. This will be fixed in my next PR. I think we should still keep it and the activity should be failure-safe and handle the case when the response is empty. What do you think?

app/src/androidTest/java/com/github/epfl/meili/poi/PoiActivityTest.kt
tberkane: Ok

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
tberkane: Yes, that's better. It's done now.

MarcelTorne: Perfect! Thanks for the work

MarcelTorne (approved):

aman3014 (approved):

PR #100, opened 2021-04-01
Image choice, upload and download to/from Firebase

aman3014:
This PR adds a general interface to allow uploading and downloading files to/from firebase storage. It also shows how Firebase and Picasso can be easily used together.
- FirebaseStorageService (implements StorageService): consists exclusively of Firebase API calls. This is why there are no tests for this service.
- PhotoDemoActivity: temporary activity to showcase the use of FirebaseStorageService and Picasso (displaying an image). Since this is temporary code, IMO we don't need to test it. The functionality can be easily tested once we integrated image display in some feature such as the forum.

This PR also removes redundant dependencies in app.gradle


app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+    private lateinit var imageView: ImageView
+    private var filePath: Uri? = null
+
+    private val getContent: ActivityResultLauncher<String> =
+        registerForActivityResult(ActivityResultContracts.GetContent()) { filePath ->
+            val bitmap = if (Build.VERSION.SDK_INT < 28) {
+                getBitmap(contentResolver, filePath)
+            } else {
+                decodeBitmap(createSource(contentResolver, filePath))
+            }
```
ahmedc3ca, line 39: Can you please explain this part? I don't understand what's the difference between build versions earlier than 28 and the ones after 28.

app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    fun onClick(view: View) {
+        when(view) {
+            choose -> getContent.launch("image/*")
+            upload -> {
+                if (filePath == null) {
+                    Toast.makeText(applicationContext, "Choose an image first", Toast.LENGTH_SHORT).show()
+                } else {
+                    storageService().uploadFile("images/myfavimage", filePath!!, {
```
ahmedc3ca, line 66: Since this is a demo activity I understand the usage of "images/myfavImage" as a path.
But just to be sure for later uses, let's say for instance I want the forum to store the image; do I only need to change the path or will I also need to modify the storageService?
Because for that case the forum path doesn't store only images it stores the entire forum object.

ahmedc3ca (commented):
A great demonstration of how to use an external API to simplify your life :P . I just have a few questions that I left in the review.

app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
aman3014: Yes! In fact, the `getBitmap` function has been deprecated and replaced by `decodeBitmap` in Android. However, SDK versions < 28 don't support `decodeBitmap`. Since our app's min SDK version is 23, we need to support both old and new versions.

app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
aman3014: You only need to change the path, no modifications to the storageService are necessary.

As for the forum example, one possible path is "images/forum/{post_id}". For both upload and download, this path will uniquely identify the image associated to the forum post.

Another example we could consider is profile pictures. In this case, we could, for example, use "images/avatar/{user_id}" as the path.

Hope this makes sense!

Great question! I will add this explanation as documentation too.

app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.appcompat.app.AppCompatActivity
+import com.github.epfl.meili.storage.FirebaseStorageService
+import com.github.epfl.meili.storage.StorageService
+import com.squareup.picasso.Picasso
+
+class PhotoDemoActivity : AppCompatActivity() {
+
+    companion object {
+        var storageService: () -> StorageService = { FirebaseStorageService }
```
tberkane, line 23: Wow! Cool trick!

app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
```diff
@@ -0,0 +1,80 @@
...
+    private lateinit var upload: Button
+    private lateinit var show: Button
+    private lateinit var imageView: ImageView
+    private var filePath: Uri? = null
+
+    /*
+        the getBitmap function has been deprecated and replaced by decodeBitmap in Android.
+        However, SDK versions < 28 don't support decodeBitmap.
+        Since Meili's min SDK version is 23, we need to support both old and new versions.
+    */
```
tberkane, line 36: Nitpick: move this comment inside of the method because like this it looks like documentation to me

app/src/main/java/com/github/epfl/meili/storage/FirebaseStorageService.kt
```diff
@@ -0,0 +1,21 @@
+package com.github.epfl.meili.storage
+
+import android.net.Uri
+import com.google.firebase.storage.FirebaseStorage
+
+object FirebaseStorageService: StorageService {
+
+    private val firebaseStorage = FirebaseStorage.getInstance()
+
+    override fun uploadFile(remotePath: String, filePath: Uri, onSuccessCallback: () -> Unit, onFailureCallback: () -> Unit) {
```
tberkane, line 10: Shouldn't we compress the image before storing it in firebase?

tberkane (commented):
Looks great! I am just concerned about not compressing the pictures before uploading them, which will cause trouble to our app once it gets more and more users.
And I agree that testing the temporary activity is not necessary.
Good job!

app/src/main/java/com/github/epfl/meili/storage/FirebaseStorageService.kt
aman3014: Good idea, done!

app/src/main/java/com/github/epfl/meili/PhotoDemoActivity.kt
aman3014: Fixed

ahmedc3ca (approved):

tberkane (approved):

MarcelTorne (approved):
Everything seems good to me! Thanks for the work!

dyxcooh (commented):
It's all good for me! I like the fact that you are putting more comments in the code

PR #101, opened 2021-04-02
Cleanup & Refactor

aman3014:
- Removes register login and latest message activities related files
- Address most build warnings and cleanup code


app/src/main/java/com/github/epfl/meili/home/GoogleSignInActivity.kt
```diff
@@ -28,7 +28,7 @@ class GoogleSignInActivity : AppCompatActivity() {
         updateUI()
     }
 
-    fun onGoogleButtonClick(view: View) {
+    fun onGoogleButtonClick(@Suppress("UNUSED_PARAMETER") view: View) {
```
ahmedc3ca, line 31: out of curiosity what does @Supress("UNUSED_PARAMETER") do?

ahmedc3ca (approved):
It's unfortunate that we had to delete this many files but it's for the best.

app/src/main/java/com/github/epfl/meili/home/GoogleSignInActivity.kt
aman3014: Without this annotation the build generates an unused parameter warning. Since we can't remove this from the function declaration as it's a onClick listener, this is the only way to not have this warning

MarcelTorne (approved):
Looks good to me! Thanks a lot for the work!

PR #105, opened 2021-04-06
Fix bug which made author of post not be shown

Bwananah:


aman3014 (approved):
Thanks for fixing this so quickly.

ahmedc3ca (approved):
This commit is almost equivalent to the one I have approved before so it's fine by me.

Bwananah (commented):
Ok for me, can be merged. Thanks for the changes

PR #106, opened 2021-04-11
Feature/photo

tberkane:
Pull request for camera, which can:
- take photos with button/lower volume button
- zoom on pinch
- focus on press
- rotate
- display photo after it is taken


app/src/main/java/com/github/epfl/meili/MainActivity.kt
```diff
@@ -1,16 +1,19 @@
...
 import androidx.appcompat.app.AppCompatActivity
 import com.github.epfl.meili.forum.ForumActivity
 import com.github.epfl.meili.home.GoogleSignInActivity
 import com.github.epfl.meili.map.MapActivity
+import com.github.epfl.meili.photo.CameraActivity
 import com.github.epfl.meili.messages.ChatLogActivity
 import com.github.epfl.meili.review.ReviewsActivity
 import com.google.android.gms.maps.model.LatLng
 import com.google.android.gms.maps.model.PointOfInterest
+import java.io.File
```
Bwananah, line 16: Do we need this ?

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
```diff
@@ -0,0 +1,283 @@
...
+            ImageCapture.OutputFileOptions.Builder(photoFile).build(),
+            ContextCompat.getMainExecutor(this),
+            object :
+                ImageCapture.OnImageSavedCallback { // object with callback functions for pictures
+                override fun onError(exc: ImageCaptureException) {
+                    Log.e(TAG, "Photo capture failed: ${exc.message}", exc)
+                }
+                override fun onImageSaved(output: ImageCapture.OutputFileResults) {
+                    val intent = Intent(applicationContext, PhotoDisplayActivity::class.java)
+                    intent.putExtra(URI_KEY, Uri.fromFile(photoFile))
```
Bwananah, line 194: Why do you save as a file instead of passing the photo directly ?

Bwananah (commented):
Good job ! I made a few comments for stuff I would like to understand.

app/src/main/java/com/github/epfl/meili/MainActivity.kt
tberkane: Good catch, I removed it

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
tberkane: Yes, it would be better to only save the photo (like WhatsApp, Telegram, ... do) once it has been confirmed that the photo will be posted. Or do you think we should not save it at all?

Bwananah: Not sure, we can leave it as it is for now and see with Aman what the best way is for posting photos

Bwananah (approved):

app/build.gradle
```diff
@@ -97,6 +97,20 @@ dependencies {
...
+    // CameraX core library using the camera2 implementation
+    def camerax_version = "1.0.0-rc03"
+    // The following line is optional, as the core library is included indirectly by camera-camera2
+    implementation "androidx.camera:camera-core:${camerax_version}"
+    implementation "androidx.camera:camera-camera2:${camerax_version}"
+    // If you want to additionally use the CameraX Lifecycle library
+    implementation "androidx.camera:camera-lifecycle:${camerax_version}"
+    // If you want to additionally use the CameraX View class
+    implementation "androidx.camera:camera-view:1.0.0-alpha22"
+    // If you want to additionally use the CameraX Extensions library
```
aman3014, line 111: Don't think this and above comments are necessary

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
aman3014: Yes, I can take care of integrating this with the upload in my next sprint.

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
```diff
@@ -0,0 +1,251 @@
...
+
+            // Attach the lifecycle of the camera to the activity's lifecycle
+            camera = cameraProvider.bindToLifecycle(
+                this as LifecycleOwner, cameraSelector, preview, imageCapture
+            )
+
+        }, ContextCompat.getMainExecutor(this))
+    }
+
+    /** Takes a picture */
```
aman3014, line 158: Just a little nitpick: while comments are nice, obvious ones just clutter the code imo. It would be great if you could this and other such comments in this file!

aman3014 (commented):
Thanks for your work guys!

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
tberkane: Done

aman3014 (approved):

aman3014:
We can't apply filters to photos right? Should we unlink #59

PR #116, opened 2021-04-13
Display POIs on Map as Personalised Markers

MarcelTorne:


aman3014 (approved):
Looks good, thanks!

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
```diff
@@ -58,20 +71,86 @@ class MapActivity : AppCompatActivity(), OnMapReadyCallback, GoogleMap.OnPoiClic
...
+        poiMarkerViewModel.mPointsOfInterestStatus.observe(this) {
+            addItems(it)
+        }
+    }
+
+    private fun addItems(map: Map<String, PoiMarkerViewModel.PointOfInterestStatus>) {
+        val newMap = HashMap<PoiItem, PoiMarkerViewModel.PointOfInterestStatus>()
+        for (entry in map.entries) {
+            val poiItem: PoiItem
+            if (poiItems.containsKey(entry.key)) {
```
tberkane, line 122: You could use the getOrDefault method of map I think

app/src/main/java/com/github/epfl/meili/map/PoiRenderer.kt
```diff
@@ -0,0 +1,50 @@
...
+    : DefaultClusterRenderer<PoiItem>(context, map, clusterManager) {
+
+    private var poiStatusMap: Map<PoiItem, PoiMarkerViewModel.PointOfInterestStatus>? = null
+
+    override fun onBeforeClusterItemRendered(item: PoiItem, markerOptions: MarkerOptions) {
+        val icon: BitmapDescriptor
+        if (poiStatusMap == null || !poiStatusMap!!.contains(item)) {
+            icon = DEFAULT_ICON
+        } else {
+            when (poiStatusMap!![item]) {
```
tberkane, line 24: Nitpick: lift the icon assignment out of the when

tberkane (commented):
Great job!

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
MarcelTorne: Unfortunately it's not possible with this APK version, because poiItems is a Map instance and not HashMap

app/src/main/java/com/github/epfl/meili/map/PoiRenderer.kt
MarcelTorne: Awesome! Didn't know about this. Thanks!

app/src/main/java/com/github/epfl/meili/map/MapActivity.kt
tberkane: Cool :+1:

tberkane (approved):

PR #117, opened 2021-04-14
refactor reviews to use general db

aman3014:


app/src/main/java/com/github/epfl/meili/database/Database.kt
```diff
@@ -1,11 +1,13 @@
...
+import kotlin.collections.HashSet
 
-abstract class Database<T>(private val path: String): Observable() {
-    abstract var values: Map<String, T>
 
-    abstract fun addElement(uid: String?, element: T?)
+abstract class Database<T>(private val key: String): Observable() {
+    abstract var elements: Map<String, T>
+
+    abstract fun addElement(uid: String, element: T?)
```
MarcelTorne, line 10: @aman3014 really cool that you managed to make it work without the nullable for uid, is it possible to do the same for element?

app/src/main/java/com/github/epfl/meili/database/Database.kt
```diff
@@ -1,11 +1,13 @@
 package com.github.epfl.meili.database
 
 import java.util.*
+import kotlin.collections.HashSet
 
-abstract class Database<T>(private val path: String): Observable() {
-    abstract var values: Map<String, T>
 
-    abstract fun addElement(uid: String?, element: T?)
+abstract class Database<T>(private val key: String): Observable() {
```
MarcelTorne, line 7: Personally, I believe that `path` as variable name  was clearer than `key`. What do you think? Did you have a strong reason to change it?

MarcelTorne (commented):
Thanks for the changes! Left two small comments.

app/src/main/java/com/github/epfl/meili/database/Database.kt
aman3014: Unfortunately it's not very easy. I read that we would have to switch over to using MockK (K for Kotlin) to do that, which can be time consuming.

app/src/main/java/com/github/epfl/meili/database/Database.kt
aman3014: Okay, you're right, since you might want to have more complex paths than simple document keys.

app/src/main/java/com/github/epfl/meili/database/Database.kt
MarcelTorne: Sounds good. Thanks!

app/src/main/java/com/github/epfl/meili/database/Database.kt
MarcelTorne: Great! Thank you!

MarcelTorne (approved):

ahmedc3ca (approved):
"database" is indeed more descriptive than "service" 👍

PR #119, opened 2021-04-16
Feature/unified UI

ahmedc3ca:
In this pull request, I linked the core features (including this sprint's features) together to look like something similar to the desired final product. I still left the main activity for later sprints so we can still add our new functionalities as a button on the main page as we used to. Everything almost works the same (i.e. almost no functionality was changed), this was mainly an interface side modification.

As usual, if any technical problems or remarks raised from this PR please leave them in the review. Any feedback on the interface itself is very welcome but won't be changed immediately but rather taken into account for the next big UI cleanup (or a smaller sooner PR).

I know it says 43 files changed but only 11 of them are code changes. The rest of them are layout changes (so things like adding background = blue) so don't worry!


MarcelTorne:
Damn! Looks really awesome! Thanks Ahmed!

However, I think we should not merge with such a low coverage. Probably just adding tests that click on buttons and verify that the intent launched is the correct one will already move the code coverage up.

MarcelTorne (changes requested):
As mentioned in the comment above, I think it would be great to have more code coverage. And I think we can achieve that by adding tests that click on buttons and verify that the launched intent is the expected one.

Thanks again for the work!

app/src/androidTest/java/com/github/epfl/meili/review/ReviewMenuButtonsTest.kt
```diff
@@ -0,0 +1,251 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito
+import org.mockito.Mockito.`when`
+
+
+@Suppress("UNCHECKED_CAST")
+@RunWith(AndroidJUnit4::class)
+class ReviewMenuButtonsTest {
```
aman3014, line 54: Is there a reason to include the private methods and all the mocks in this class? I don't see where they are being used.

aman3014 (commented):
How can I access the review, form and chat activities? When I launch, I'm only able to see sign in and map view, from which I can further access camera and images.

app/src/androidTest/java/com/github/epfl/meili/review/ReviewMenuButtonsTest.kt
ahmedc3ca: They're used to mock the forum activity when it's opened. I was getting internal firestore errors otherwise. By the way, you can access the chat, the review and the forum by clicking on each individual POI on the map (I'm replying here since I can't reply to the general review comment).

aman3014: I tried clicking on a POI and nothing happens.

ahmedc3ca: I removed the unused methods and mocks. If you think there are still some useless mocks please mark them using the multiple line comment feature.
As for the POI, that's weird. Does something happen when you click on the POI with the current main at least?

tberkane (changes requested):
The new design of the app looks really good! Great work!
I have not found a way to access the POI info view and the camera view though. So I think that those should be added.
I have not looked at the code yet.

ahmedc3ca:
> I have not found a way to access the POI info view and the camera view though. So I think that those should be added.

@tberkane cf PR #90 : "Launching by clicking on a poi on the map will come in a later pr." Not sure how we can add POI info fragment to the main UI if this feature is still not available. We can integrate it as soon as this later pr is made. Anyways, this is why I left the mainActivity in the project. It's to show these features during the live demos. For the camera, it's supposed to be on the top right when you're in the map activity let me know if that's not the case.

tberkane:
> > I have not found a way to access the POI info view and the camera view though. So I think that those should be added.
> 
> @tberkane cf PR #90 : "Launching by clicking on a poi on the map will come in a later pr." Not sure how we can add POI info fragment to the main UI if this feature is still not available. We can integrate it as soon as this later pr is made. Anyways, this is why I left the mainActivity in the project. It's to show these features during the live demos. For the camera, it's supposed to be on the top right when you're in the map activity let me know if that's not the case.

My bad, that phrase you quoted is outdated. Now when you click on a POI, the poi info activity also displays.
I don't see the camera.

ahmedc3ca:
> > > I have not found a way to access the POI info view and the camera view though. So I think that those should be added.
> > 
> > 
> > @tberkane cf PR #90 : "Launching by clicking on a poi on the map will come in a later pr." Not sure how we can add POI info fragment to the main UI if this feature is still not available. We can integrate it as soon as this later pr is made. Anyways, this is why I left the mainActivity in the project. It's to show these features during the live demos. For the camera, it's supposed to be on the top right when you're in the map activity let me know if that's not the case.
> 
> My bad, that phrase you quoted is outdated. Now when you click on a POI, the poi info activity also displays.
> I don't see the camera.

Ty for bringing up the missing camera it had disappeared due to the last conflict fix. It should reappear now.
As for the POI info, I'm sorry but I can't add it to this PR. The reason being is that since this is "a general PR that should contain all of the functionalities together", every time someone merges a feature before me I have to add this to this PR. By the time I add this new feature and reopen the PR, a new feature rises that I have to again re-add to my PR. This cycle has been going for a while and it's very time-consuming for me. Plus this PR has to be merged someday if I continue with this pattern it will never merge. I think it will be better to go with the solution I have suggested before and Just use the mainActivity for the demo if we need to show the POI info feature.

tberkane:
> > > > I have not found a way to access the POI info view and the camera view though. So I think that those should be added.
> > > 
> > > 
> > > @tberkane cf PR #90 : "Launching by clicking on a poi on the map will come in a later pr." Not sure how we can add POI info fragment to the main UI if this feature is still not available. We can integrate it as soon as this later pr is made. Anyways, this is why I left the mainActivity in the project. It's to show these features during the live demos. For the camera, it's supposed to be on the top right when you're in the map activity let me know if that's not the case.
> > 
> > 
> > My bad, that phrase you quoted is outdated. Now when you click on a POI, the poi info activity also displays.
> > I don't see the camera.
> 
> Ty for bringing up the missing camera it had disappeared due to the last conflict fix. It should reappear now.
> As for the POI info, I'm sorry but I can't add it to this PR. The reason being is that since this is "a general PR that should contain all of the functionalities together", every time someone merges a feature before me I have to add this to this PR. By the time I add this new feature and reopen the PR, a new feature rises that I have to again re-add to my PR. This cycle has been going for a while and it's very time-consuming for me. Plus this PR has to be merged someday if I continue with this pattern it will never merge. I think it will be better to go with the solution I have suggested before and Just use the mainActivity for the demo if we need to show the POI info feature.

I don't think it makes sense to merge something which removes features.
Also, I noticed that none of the features are accessible anymore without being signed in, even though everything should be able to function, except the chat.

ahmedc3ca:
> I don't think it makes sense to merge something which removes features.
> Also, I noticed that none of the features are accessible anymore without being signed in, even though everything should be able to function, except the chat.
> Also, just a question, but I noticed that the issue linked with this PR is already closed. Is this an oversight?
> Finally, would it also be possible to achieve higher coverage?
> Thanks!

That's my point exactly, we're not removing the feature since it's still accessible from the main Activity. If you want I can make a new commit that changes the android manifest. 
What other solution are you suggesting? Because I don't see any other one.
As for the sign-in, it seemed more seamless to ask the user to login in at the beginning. Again, this is not the final product to be delivered for the user, we can easily change this later.
Yes, I closed it by mistake and I can't reopen it.
For the coverage, after Marcel pointed it out, I made some tests and this is the highest coverage I can get with this (84% is not that bad).

tberkane:
> > I don't think it makes sense to merge something which removes features.
> > Also, I noticed that none of the features are accessible anymore without being signed in, even though everything should be able to function, except the chat.
> > Also, just a question, but I noticed that the issue linked with this PR is already closed. Is this an oversight?
> > Finally, would it also be possible to achieve higher coverage?
> > Thanks!
> 
> That's my point exactly, we're not removing the feature since it's still accessible from the main Activity. If you want I can make a new commit that changes the android manifest.
> What other solution are you suggesting? Because I don't see any other one.
> As for the sign-in, it seemed more seamless to ask the user to login in at the beginning. Again, this is not the final product to be delivered for the user, we can easily change this later.
> Yes, I closed it by mistake and I can't reopen it.
> For the coverage, after Marcel pointed it out, I made some tests and this is the highest coverage I can get with this (84% is not that bad).

Okay but the main activity is not accessible anymore.
I am suggesting that you add a way to access the poi info activity once a poi has been clicked.

ahmedc3ca:
> Okay but the main activity is not accessible anymore.
> I am suggesting that you add a way to access the poi info activity once a poi has been clicked.

This is why I'm suggesting to change the android manifest accordingly

aman3014:
For the sign in, the forum and the review system are both already compatible with this. They work in a limited manner for unauthenticated users. When you click on chat I guess you could redirect to the sign in activity (?) By not doing this we are actually removing features (namely unauthenticated functionality) from the app which is not ideal.

As for the linked issue, I found how to reopen it actually, there is a button at the bottom when you go to the issue page. Additionally, we should only put cards into the "Done Sprint #" board once the feature is merged. I saw that this issue is put in "Done Sprint 4" which was 2 sprints ago. I think that we should move the issue to Sprint Tasks and move it to Done Sprint # when the PR is merged.

ahmedc3ca:
I have changed the android manifest to have as main activity the old one. This way we can show the features without logging in and we can have the partially unified UI accessible from the google sign-in activity button.

app/src/main/java/com/github/epfl/meili/home/GoogleSignInActivity.kt
```diff
@@ -28,6 +29,11 @@ class GoogleSignInActivity : AppCompatActivity() {
         updateUI()
     }
 
+    fun OnMapViewButtonClick(@Suppress("UNUSED_PARAMETER") view: View) {
```
tberkane, line 32: Nitpick: function names should start with a lowercase.

tberkane (approved):
Wow, the app looks way cooler now! I especially like the logo!
Thanks for your work. I just had a small nitpick to make.

aman3014 (approved):
Thanks for your work and for addressing the comments!

app/src/main/java/com/github/epfl/meili/home/GoogleSignInActivity.kt
ahmedc3ca: Changed 👍

MarcelTorne (approved):
Really cool! Thanks for all the work and for improving code coverage!

PR #131, opened 2021-04-17
add images to forum posts

aman3014:
Adds images to the forum. Each post can now have one image associated to it. This image can be added either from the gallery or by taking a picture with the Camera. When seeing the post, the associated picture is displayed. If there isn't an associated picture, a default image (https://en.wikipedia.org/wiki/Roman_Forum#/media/File:Foro_Romano_Musei_Capitolini_Roma.jpg) is shown.

The low diff coverage is due to methods like compression, upload, and image display (using Picasso). But since these are API calls, they don't constitute that many LoCs which is why the total coverage doesn't change at all.


app/src/main/java/com/github/epfl/meili/forum/ForumActivity.kt
```diff
@@ -63,28 +93,34 @@ class ForumActivity : AppCompatActivity() {
...
             else -> openPost(view.findViewById(R.id.post_id))
         }
     }
 
     private fun openPost(view: View) {
         val postId: String = (view as TextView).text.toString()
         val intent: Intent = Intent(this, PostActivity::class.java)
-                .putExtra("Post", viewModel.getElements().value?.get(postId))
+                .putExtra(Post.TAG, viewModel.getElements().value?.get(postId))
+                .putExtra("PostId", postId)
```
tberkane, line 123: Nitpick: create a constant for this.

app/src/main/java/com/github/epfl/meili/forum/ForumActivity.kt
```diff
@@ -146,6 +186,37 @@ class ForumActivity : AppCompatActivity() {
...
+            }
+
+            runOnUiThread {
+                this.bitmap = bitmap
+                displayImageView.setImageBitmap(bitmap)
+            }
+        }
+    }
+
+    private fun compressed(bitmap: Bitmap): ByteArray {
```
tberkane, line 204: Nitpick: this method name could be a bit more descriptive

tberkane (approved):
Looks good!

Bwananah (approved):
Looks good

PR #133, opened 2021-04-19
Change from Overpass API to GooglePlaces API to fetch POIs

MarcelTorne:


app/src/main/java/com/github/epfl/meili/poi/PointOfInterest.kt
```diff
@@ -9,22 +9,26 @@ data class PointOfInterest(
...
+        return START_CHAR + "POI:" + "lat:" + latitude + "long:" + longitude + ",name:" + name + ",uid:" + uid + ",icon:" + icon + END_CHAR
     }
 
     override fun equals(other: Any?): Boolean {
-        if (other != null && other::class.java == PointOfInterest::class.java){
-            val otherPoi =  other as PointOfInterest
-            return otherPoi.latitude.equals(latitude) && otherPoi.longitude==longitude && otherPoi.name == name && otherPoi.uid == uid
+        if (other != null && other::class.java == PointOfInterest::class.java) {
+            val otherPoi = other as PointOfInterest
+            return otherPoi.latitude.equals(latitude) && otherPoi.longitude == longitude && otherPoi.name == name
```
tberkane, line 24: Nitpick: you can replace the call to equals by ==

tberkane (commented):
Great job! Now that we have the custom POIs displayed on the map, maybe we should remove those from the google maps api?

MarcelTorne:
> Great job! Now that we have the custom POIs displayed on the map, maybe we should remove those from the google maps api?

Yes I agree. I am going to ask on the team group to see if everyone is onboard with it. Thanks!

tberkane (approved):
Perfect!

aman3014 (approved):
Great work!

PR #134, opened 2021-04-22
Add PoiService that caches POI response from API

MarcelTorne:
I am adding a service that provides a 2 level cache. The first level is the data saved in the object and the second level is the data saved locally on the phone. If there is no data valid in any of the two levels then it fetches from the Google Places API.


Bwananah (approved):
Good job ! Looks good !

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
```diff
@@ -0,0 +1,112 @@
...
+        val latLng: PoiLocation? = null
+)
+
+data class PoiLocation(
+        @SerializedName("lat")
+        val latitude: Double? = null,
+        @SerializedName("lng")
+        val longitude: Double? = null
+)
+
```
aman3014, line 108: Can't we use LatLng (https://developers.google.com/android/reference/com/google/android/gms/maps/model/LatLng) for this?

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
```diff
@@ -0,0 +1,112 @@
...
+data class PoiOpeningHours(
+        @SerializedName("open_now")
+        val openNow: Boolean? = null
+)
```
aman3014, line 112: Same comment

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
```diff
@@ -0,0 +1,112 @@
...
+            return PointOfInterest(geometry!!.latLng!!.latitude!!, geometry.latLng!!.longitude!!, name!!, uid!!, icon!!, poiTypes!!, openingHours.openNow)
+        }
+        return PointOfInterest(geometry!!.latLng!!.latitude!!, geometry.latLng!!.longitude!!, name!!, uid!!, icon!!, poiTypes!!)
+    }
+}
+
+data class PoiGeometry(
+        @SerializedName("location")
+        val latLng: PoiLocation? = null
+)
```
aman3014, line 100: Same comment about single attribute class here.

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
```diff
@@ -0,0 +1,112 @@
...
+        private const val GOOGLE_PLACES_URL = "https://maps.googleapis.com/maps/api/place/nearbysearch/json?"
+        private const val MAX_COVERAGE_RADIUS = 3000
+        private const val TAG = "PoiGoogleRetriever"
+    }
+}
+
+data class GooglePlacesResponse(
+        @SerializedName("results")
+        val pointsOfInterest: List<PlacesPointOfInterest> = ArrayList()
+) {
```
aman3014, line 62: Is there a strong reason for the existence of this class? A class with a single attribute looks unnecessary to me.

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
MarcelTorne: Thanks for your comments. The thing is that this class in only used for the json deserialisation, so if I put LatLng it won't deserialise the object correctly and instead I have to get each value separately (latitude and longitude).

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
MarcelTorne: Again, I am only using these classes for the json object deserialisation, so this depends on how the object has been formed. Moreover, it is possible that for some of these classes (not GooglePlacesResponse but PoiOpeningHours) that the Json provides more attributes. However, they didn't seem important for the purpose of our application and hence I am not retrieving them.

MarcelTorne: I am going to add a comment specifying that these data classes are only for deserialisation (after fetching the info the answer is transformed into a List<PointOfInterest>).

aman3014:
@MarcelTorne is it possible to have a general cache service which the poi service cached could inherit from ?

MarcelTorne:
> @MarcelTorne is it possible to have a general cache service which the poi service cached could inherit from ?

@aman3014 I suppose it has to be possible, however, there are some really specific things related to the validity of the cached value that I am not sure how we could generalise. For example, in this case a cached value is valid depending on time (which could be generalised) but also on the distance from the last request (which makes this generalisation harder). Do you have any ideas?

aman3014:
I see. Something I could think of was having a general cache which only checks the time validity (since that is likely to be the only criterion for other caches) whereas the poi service cache checks for the distance in addition (and then call the super.valid() method or equivalent)

MarcelTorne:
> I see. Something I could think of was having a general cache which only checks the time validity (since that is likely to be the only criterion for other caches) whereas the poi service cache checks for the distance in addition (and then call the super.valid() method or equivalent)

I see, good idea! Do you think we can merge this PR, and will create a new task for next week that consists in generalising the cache service? Since I am already off the time estimate.

aman3014 (approved):
Thanks a lot for the work! We should make sure to generalise the cache service in the next sprint.

PR #136, opened 2021-04-22
Refactor Menu Inflators for more maintainability

ahmedc3ca:
Created object MenuInflaterHelper to decrease duplicate code


app/src/main/java/com/github/epfl/meili/util/MenuInflaterHelper.kt
```diff
@@ -0,0 +1,61 @@
...
+                    .putExtra(MapActivity.POI_KEY, poi)
+            }
+            R.id.menu_forum_from_review-> {
+                Intent(acitvity, ForumActivity::class.java)
+                    .putExtra(MapActivity.POI_KEY, poi)
+            }
+            else -> {
+                Intent(acitvity, ForumActivity::class.java)
+            }
+        }
```
aman3014, line 47: I think we could replace menu_chat_from_review and menu_chat_from_forum with buttons with the same id menu_chat. I checked and we can indeed have buttons with the same id as long as they are in different layouts. This will remove the duplication in this part of the code. What do you think ?

aman3014: Sorry, my comment was on outdated code. What you did already looks better! But I think we could still do this, for the sake of having the same id everywhere.

aman3014: Also, I think we should not instantiate the activity but rather store launchedActivity::class.java directly. For eg. ReviewsActivity::class.java

ahmedc3ca: Both done 👍 I wasn't sure about the buttons having the same name but you're right it works

tberkane (approved):
Looks good!

aman3014 (approved):
Thanks!

PR #147, opened 2021-04-23
Friend nearby users using Nearby API

aman3014:
### Allows two users to become friends using the Nearby API.

Both users just need to click on the "Find my Friend" button. Their phone's Bluetooth and WIFI will automatically be turned on and off as required, leaving them in the initial state. The devices will discover each other and a prompt will be displayed to both users.

"Accept friendship with {other_username}" 

"Confirm friendship by verifying that the same code is displayed on both devices: {code}". This is to make sure that the connection is authenticated and that the users don't become friends with random people.

Once they both accept, their friendship will be stored to Firestore (or will be scheduled to be stored when the devices connect to the internet, since internet connectivity is not a requirement for the above connection procedure).

### Speed up transition from PhotoEdit to Forum. Also refactor PhotoEditActivity tests so that they pass more consistently on Cirrus.

The low coverage is mainly due to the added functionality in LocationService - permission request and location enable prompt - being hard to test.

The core functionalities of the NearbyActivity are well tested using mocks. However, the handling of location permission permission isn't covered by the tests.


MarcelTorne (approved):
Thanks for making this cool feature!

Bwananah (approved):
Good job !

PR #149, opened 2021-04-25
On POI request error, schedule new request after some time

MarcelTorne:


aman3014 (commented):
Looks good! I was just thinking that 5 minutes for the TIME_UNTIL_NEXT_REQUEST might be too long and the user would not wait such a long time. What do you think about re-requesting instantly and having a limit on the number of re-requests (for eg. 2). Because if two requests are not successful there likely is a problem and subsequent requests probably won't be successful either)

tberkane (approved):

MarcelTorne:
> Looks good! I was just thinking that 5 minutes for the TIME_UNTIL_NEXT_REQUEST might be too long and the user would not wait such a long time. What do you think about re-requesting instantly and having a limit on the number of re-requests (for eg. 2). Because if two requests are not successful there likely is a problem and subsequent requests probably won't be successful either)

Yes I see, I think it is a good idea to add a couple of immediate requests but I think we should also keep the 5 minutes one (or different time) because when there's no internet connection and no data on your device then you receive an error but you might want to retry again since your internet connection can change. What do you think?

aman3014 (approved):
Thanks for the changes!

PR #151, opened 2021-04-26
Create general Cache Service

MarcelTorne:
Refactoring code from PoiCachedService into a general cache service that can be reused by different activities and classes. I am also creating an interface `ResponseFetcher` which is the service used by the cache service to get the response in case it is not found in the local cache.


app/src/main/java/com/github/epfl/meili/cache/CacheService.kt
```diff
@@ -0,0 +1,156 @@
...
+    protected open fun isCacheValid(cachedTimestamp: Long): Boolean {
+        return verifyValidity(cachedTimestamp)
+    }
+
+    /**
+     * Data is considered valid if it was retrieved less than 1 hour ago and the new request comes
+     * from less than 1km far away from the last data
+     *
+     * @return whether the data in the object is valid or not
+     */
```
aman3014, line 139: Leftover comment

app/src/main/java/com/github/epfl/meili/cache/CacheService.kt
```diff
@@ -0,0 +1,156 @@
...
+    private fun retrieveTimeOfResponse(): Long {
+        return mPrefs.getLong(TIMESTAMP_KEY, 0L)
+    }
+
+    /**
+     * Data is considered valid if it was retrieved less than 1 hour ago and the new request comes
+     * from less than 1km far away from the last data
+     *
+     * @return whether the data saved in shared preferences is valid or not
+     */
```
aman3014, line 129: Leftover comment

aman3014 (approved):
Looks great, thanks!

app/src/main/java/com/github/epfl/meili/cache/CacheService.kt
MarcelTorne: Changed! Thanks for pointing it out!

Bwananah (approved):
Very clean, good job !

PR #154, opened 2021-04-28
Switch between front and back cameras

tberkane:


Bwananah (approved):
Looks fine by me ! Well done !

app/src/androidTest/java/com/github/epfl/meili/photo/PhotoEditActivityTest.kt
```diff
@@ -142,6 +142,5 @@ class PhotoEditActivityTest {
             )
         )
         fab.check(ViewAssertions.matches(isDisplayed()))
-        fab.perform(click())
```
aman3014, line 145: Why is this being removed?

app/src/androidTest/java/com/github/epfl/meili/photo/CameraActivityTest.kt
```diff
@@ -69,7 +69,9 @@ class CameraActivityTest {
     @Test
     fun takePhotoWithVolumeDownTest() {
         val mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        Thread.sleep(1000)
         mDevice.pressKeyCode(KeyEvent.KEYCODE_VOLUME_DOWN)
+        Thread.sleep(1000)
```
aman3014, line 74: Why are these sleeps being added, is the CameraActivity slower now?

tberkane: They are added so that the tests pass.
This is a test so it has no impact on the camera activity.

app/src/androidTest/java/com/github/epfl/meili/photo/PhotoEditActivityTest.kt
tberkane: Because it is what made the tests fail on cirrus

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
```diff
@@ -36,14 +35,17 @@ class CameraActivity : AppCompatActivity() {
     private var imageCapture: ImageCapture? = null // is null when camera hasn't started
     private lateinit var cameraProvider: ProcessCameraProvider
     private lateinit var preview: Preview
-    private lateinit var cameraSelector: CameraSelector
     private lateinit var camera: Camera
 
     private lateinit var outputDirectory: File // directory where photos get saved
 
     private lateinit var cameraButton: ImageButton
+    private lateinit var switchCameraButton: ImageButton
```
dyxcooh, line 43: Why do you use lateinit var here?

tberkane: Because I initialize it in onCreate

app/src/androidTest/java/com/github/epfl/meili/photo/CameraActivityTest.kt
aman3014: What I meant was that the tests were passing without this before, so the fact that we need to add this now implies that the camera activity is slower and we need to wait more for the activity to get ready. Is that correct?

AFAIK the problem is in the photo edit tests, not the camera ones.

app/src/androidTest/java/com/github/epfl/meili/photo/PhotoEditActivityTest.kt
aman3014: This should be fixed in #147, since clicking on fab now doesn't pass through CameraActivity, the intent goes directly to the caller of Camera Activity

tberkane: Perfect, thanks!

app/src/androidTest/java/com/github/epfl/meili/photo/CameraActivityTest.kt
tberkane: They weren't always passing on cirrus

app/src/main/java/com/github/epfl/meili/photo/CameraActivity.kt
```diff
@@ -139,67 +249,30 @@ class CameraActivity : AppCompatActivity() {
...
-            preview.setSurfaceProvider(previewView.surfaceProvider)
-
-            // Attach the lifecycle of the camera to the activity's lifecycle
-            camera = cameraProvider.bindToLifecycle(
-                this as LifecycleOwner, cameraSelector, preview, imageCapture
-            )
-
-        }, ContextCompat.getMainExecutor(this))
+    /** Only enable button to switch cameras if both the cameras are available */
+    private fun updateCameraSwitchButton() {
```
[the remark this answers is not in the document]
tberkane: Done

dyxcooh (approved):
Great job! Thanks for the quick reply

PR #158, opened 2021-04-30
forum posts can be down/upvoted

ahmedc3ca:


aman3014:
I think we should use `ForumViewModel` for the Forum, which can be created by extending from `MeiliViewModel`.
The `PostViewModel` is for the `PostActivity`.

ahmedc3ca:
Thank you Aman for your comment! Are you talking about just renaming the class itself or is there an already existing ForumViewModel? As for the inheritance, I totally agree that ForumViewModel should inherit from MeiliViewModel. The problem is when I tried to do it, when I initialize the database, the function call in FirestoreDatabase line 49 _this.notify observers()_ calls the update for the view models and throws an error that the database has not been initialized. Maybe you can guide me as to why this is happening since you wrote the code for the MeiliViewModel.

aman3014:
`viewModel = ViewModelProvider(this).get(MeiliViewModel::class.java) as MeiliViewModel<Post>
  viewModel.setDatabase(FirestoreDatabase("forum/$poiKey/posts", Post::class.java))`

On main, this is how the Forum Activity initialises the view model and the database. Here, the `this.notifyObservers()` call only happens once the view model's database attribute has been initialised and the problem you're describing doesn't occur.

Where did you encounter this issue? How were you initialising the view model and the database(s) when you had this problem?

ahmedc3ca:
I understand this.

If you're free, we can discuss the problem privately because it is too complicated to explain here (maybe after the stand-up?)

app/src/main/java/com/github/epfl/meili/database/AtomicPostFirestoreDatabase.kt
```diff
@@ -0,0 +1,31 @@
...
+            if (voters.contains(uid)) {
+                voters.remove(uid)
+            } else {
+                if(otherVoters.contains(uid)) otherVoters.remove(uid)
+                voters.add(uid)
+            }
+            //update to new upvoters list
+            transaction.update(sfDocRef, fieldName, voters)
+            transaction.update(sfDocRef, otherFieldName, otherVoters)
+        }
```
aman3014, line 28: I think we could increase the coverage of this file by giving a mock transaction when runTransaction is called. This should lead to all these lines being executed.

aman3014 (changes requested):
Everything looks great! I just left a suggestion for how to increase the coverage.

app/src/main/java/com/github/epfl/meili/util/MeiliViewModel.kt
```diff
@@ -6,13 +6,13 @@ import androidx.lifecycle.ViewModel
 import com.github.epfl.meili.database.Database
 import java.util.*
 
-class MeiliViewModel<T>: ViewModel(), Observer {
+open class MeiliViewModel<T>: ViewModel(), Observer {
 
     private val mElements: MutableLiveData<Map<String, T>> = MutableLiveData()
 
-    private lateinit var database: Database<T>
+    protected lateinit var database: Database<T>
```
MarcelTorne, line 13: One question, why are this changes being made?

MarcelTorne (commented):
Thank you for your work Ahmed! Everything looks good to me, I think it would be great to try to add the suggestions from @aman3014  to bring up a little bit the code coverage. I also left one small question. Thanks again for your good work!

app/src/main/java/com/github/epfl/meili/util/MeiliViewModel.kt
ahmedc3ca: We did it so that the ForumViewModel can access it. Otherwise, we can't define the upvote/downvote functions in ForumViewModel.kt. Plus, I believe that the children of this class should be able to use this field since the added functionality would almost always be related to the database. Thank you for the question!

MarcelTorne: Awesome! Thank you for your answer @ahmedc3ca !

aman3014 (approved):
I've added transaction testing using an argument captor to bump the coverage up.
Thanks for this cool feature 😄

MarcelTorne (approved):
Awesome work! Thanks!

PR #179, opened 2021-05-01
Create UI for friends list

MarcelTorne:


app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.github.epfl.meili.home.Auth
+import com.github.epfl.meili.models.Friend
+import com.github.epfl.meili.nearby.NearbyActivity
+import com.github.epfl.meili.util.MeiliViewModel
+import com.github.epfl.meili.util.TopSpacingItemDecoration
+
+class FriendsListActivity : AppCompatActivity() {
+    companion object {
+        private const val FRIENDS_PADDING: Int = 15
+    }
```
ahmedc3ca, line 23: Optional: you can add
```suggestion
    companion object {
        private const val FRIENDS_PADDING: Int = 15
        private const val TAG: String = "FriendListActivity"
        private const val TITLE: String = "My Friends"     
    }
```
And change the rest of the class accordingly.

ahmedc3ca (commented):
Thank you Marcel for the work! It seems like the feature works as intended. Apart from a nitpick and a question I think this should be part of the Main :)

app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        initViewModel()
+        initRecyclerView()
+
+        supportActionBar?.title = "My Friends"
+    }
+
+    private fun initViewModel() {
+        if (Auth.getCurrentUser() == null) {
+            if (BuildConfig.DEBUG) {
```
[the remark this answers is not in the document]
aman3014: This is the same as doing an `assert` in old java, which is now deprecated. BuildConfig.DEBUG is true when running tests but false when running the app.

app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListRecyclerAdapter.kt
```diff
@@ -27,8 +27,10 @@ class FriendsListRecyclerAdapter : RecyclerView.Adapter<RecyclerView.ViewHolder>
     class FriendViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
         private val name: TextView = itemView.findViewById(R.id.friend_name)
 
+        // TODO: fetch and display Friend's info from Firestore: users/{friend.uid}
```
aman3014, line 30: I just added this TODO but we can address it in another PR, since it will (probably) take a significant amount of time.

aman3014 (approved):
Thanks for this feature!

app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListRecyclerAdapter.kt
aman3014: The name that is currently displayed is actually the uid.

app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListActivity.kt
```diff
@@ -37,20 +37,17 @@ class FriendsListActivity : AppCompatActivity() {
     }
 
     private fun initViewModel() {
```
MarcelTorne, line 41: @aman3014 I think it was better before, since now if BuildConfig.DEBUG is false then you will be continuing even if the user is null, which I think is a mistake

app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListRecyclerAdapter.kt
MarcelTorne: Great! Thank you!

app/src/main/java/com/github/epfl/meili/profile/friends/FriendsListActivity.kt
aman3014: If BuildConfig.DEBUG is false then user should not be null, otherwise it's a bug.

MarcelTorne: Yes it's a bug but I guess it can happen no? And we shouldn't be executing the rest of the code

aman3014: The point is to not have any bugs on the main. This assert is there to prevent other developers from inadvertently making a mistake and creating the friend list activity without signing in.

aman3014: The current code will just suppress this error in production.

aman3014: If you write assert(something) in Android Studio, it suggests you to convert it to

`if (BuildConfig.DEBUG && something)
    error(something violated)
`

MarcelTorne: Okay sounds good!

MarcelTorne: It's changed!

ahmedc3ca (approved):

PR #185, opened 2021-05-06
More uniform color scheme in night mode

ahmedc3ca:
Hi team! Feel free to move around the activities in the dark mode theme and tell me how it looks like :). If you feel like a color doesn't fit just tell me and I'll change it.


MarcelTorne (changes requested):
Thanks for the work Ahmed! Awesome to have night mode in the app. Do you think you could implement dark mode for the map too? I think it would be great. Here is a 5 min video that would probably help you (https://www.youtube.com/watch?v=sLZxAB1el6w)

Bwananah (approved):
Looks cool ! +1 for the map, but this is already enough of a feature imo

ahmedc3ca:
Thanks for the suggestion, Marcel! I have made a new commit with this feature 👍

MarcelTorne (approved):
Really cool! Thanks for the work and the changes @ahmedc3ca!

PR #188, opened 2021-05-06
Feature/favorite POIs

tberkane:
Adds the possibility to add pois to favorites by clicking on the star in the forum. The poi will then appear in the favourite pois in the profile


app/src/androidTest/java/com/github/epfl/meili/messages/ChatLogMenuButtonsTest.kt
```diff
@@ -134,15 +134,17 @@ class ChatLogMenuButtonsTest {
...
+//
+//    @Test
+//    fun clickReviewMenuButton() {
+//
+//        onView(
+//            allOf(
+//                withId(R.id.menu_reviews), withText("Reviews"),
+//            )
+//        ).perform(click())
+//    }
```
aman3014, line 147: This test has been passing on main, it would be weird if it stopped passing. Perhaps it's some other test which fails or causes a time out and this one is shown to be not passing?

aman3014 (commented):
Could you please define "POIs which were interacted with"? Does it require the user to post to the forum? or chat?

If that's the case, we should discuss with the team if that's really how we want to define the poi history. IMO it would be nice to add a POI to the history as soon as the user was in the vicinity of the POI, as stated in #172.

app/src/main/java/com/github/epfl/meili/profile/poihistory/PoiHistoryActivity.kt
```diff
@@ -0,0 +1,96 @@
...
+
+
+class PoiHistoryActivity : AppCompatActivity() {
+    companion object {
+        private const val CARD_PADDING: Int = 30
+        private const val ACTIVITY_TITLE = "POI History"
+
+        fun addPoiToHistory(userKey: String, poi: PointOfInterest) {
+            FirestoreDatabase( // add to poi history
+                "poi-history/$userKey/poi-history",
```
MarcelTorne, line 30: I think we should try to not hardcode the paths and maybe provide a function that will get you the path given the user key. (c.f. the code review done by the assistants

MarcelTorne (changes requested):
Thank you Thomas for this good job! I just added one nitpick and I am also waiting for your response to @aman3014 's comment on the failing test before approving.

MarcelTorne:
> Could you please define "POIs which were interacted with"? Does it require the user to post to the forum? or chat?
> 
> If that's the case, we should discuss with the team if that's really how we want to define the poi history. IMO it would be nice to add a POI to the history as soon as the user was in the vicinity of the POI, as stated in #172 (which should be linked to the PR btw)

What I think would be the best is that you add the POI once the user has the POI reachable (ie he/she is in the vicinity) and the user clicks on its marker.

app/src/androidTest/java/com/github/epfl/meili/messages/ChatLogMenuButtonsTest.kt
MarcelTorne: This has happened to me before. @tberkane did you try running it on your computer with `./gradlew connectedCheck` command? Cause sometimes the tests fail after some time and as @aman3014 has said, the error appears on another one.

tberkane:
> > Could you please define "POIs which were interacted with"? Does it require the user to post to the forum? or chat?
> > If that's the case, we should discuss with the team if that's really how we want to define the poi history. IMO it would be nice to add a POI to the history as soon as the user was in the vicinity of the POI, as stated in #172 (which should be linked to the PR btw)
> 
> What I think would be the best is that you add the POI once the user has the POI reachable (ie he/she is in the vicinity) and the user clicks on its marker. What do you think @tberkane and @aman3014 ?

Yeah I asked on whatsapp

MarcelTorne (commented):
I just realised that you are doing what had been already done by me in the PoiMarkerViewModel part. The differences are that you don't have the timestamp there (which could maybe be added) and then there is the functionality of when are we setting it to visited that I thought was what you wanted to change. Maybe you can send me a message and discuss this offline, otherwise check `MapAcitivityViewModel` class and the attributes `mPointsOfInterest` and `mPointsOfInterestStatus`

app/src/main/java/com/github/epfl/meili/models/VisitedPointOfInterest.kt
```diff
@@ -0,0 +1,9 @@
+package com.github.epfl.meili.models
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+import java.util.*
+
+@Parcelize
+data class VisitedPointOfInterest(var poi: PointOfInterest? = null, var dateVisited: Date? = Calendar.getInstance().time) :
```
MarcelTorne, line 8: @tberkane I am unsure but do you think it would be best to change the name of the class to FavouritePointOfInterest? Maybe not, just wondering.

app/src/androidTest/java/com/github/epfl/meili/photo/CameraActivityTest.kt
```diff
@@ -107,7 +107,7 @@ class CameraActivityTest {
             }
         }
 
-        onView(withId(R.id.camera_preview)).perform(touches)
+        //onView(withId(R.id.camera_preview)).perform(touches)
```
MarcelTorne, line 110: Is this line supposed to be commented?

MarcelTorne (commented):
Thanks for your work Thomas! Looks really good, just added two questions. But I also wanted to put a general remark that I think would be great to have, would it be possible to have the star button as a toggle? So that it is yellow when it is favourite and black when not. Then if you on it when it's yellow then it becomes black and is also removed from your favourite pois list. What do you think? Maybe we can add it as a task for next sprint

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisActivity.kt
```diff
@@ -0,0 +1,97 @@
...
+import com.github.epfl.meili.util.MeiliViewModel
+import com.github.epfl.meili.util.TopSpacingItemDecoration
+import java.util.*
+
+
+class FavoritePoisActivity : AppCompatActivity() {
+    companion object {
+        private const val CARD_PADDING: Int = 30
+        private const val ACTIVITY_TITLE = "Favorite POIs"
+        private const val DB_PATH = "poi-history/%s/poi-history"
```
MarcelTorne, line 27: Sorry, I forgot to mention it in the review but do you think we could change the path to something like `favourite-poi...`?

tberkane:
> Thanks for your work Thomas! Looks really good, just added two questions. But I also wanted to put a general remark that I think would be great to have, would it be possible to have the star button as a toggle? So that it is yellow when it is favourite and black when not. Then if you on it when it's yellow then it becomes black and is also removed from your favourite pois list. What do you think? Maybe we can add it as a task for next sprint

Yes good idea

app/src/androidTest/java/com/github/epfl/meili/photo/CameraActivityTest.kt
tberkane: Yes for me this test also fails sometimes on main branch, so I thinks it would be better to add a task to fix it.

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisActivity.kt
```diff
@@ -0,0 +1,97 @@
...
+    companion object {
+        private const val CARD_PADDING: Int = 30
+        private const val ACTIVITY_TITLE = "Favorite POIs"
+        private const val DB_PATH = "poi-favorite/%s/poi-favorite"
+
+        fun addPoiToFavorites(userKey: String, poi: PointOfInterest) {
+            FirestoreDatabase( // add to poi favorites
+                String.format(DB_PATH, userKey),
+                FavoritePointOfInterest::class.java
+            ).addElement(poi.uid, FavoritePointOfInterest(poi))
```
aman3014, line 33: This creates a FirestoreDatabase every time a POI is added, which is not great IMO. Instead, the database should be instantiated in the view model of the activity where the database is used (here, Forum).

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisRecyclerAdapter.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int) {
+        val pvh = (holder as PoiViewHolder)
+        pvh.bind(items[position])
+    }
+
+
+    class PoiViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+        private val poiName: TextView = itemView.findViewById(R.id.poi_name)
+        private val visitedDate: TextView = itemView.findViewById(R.id.visited_date)
```
aman3014, line 29: should we have a visited date for favourite pois? The user could potentially favourite a POI that they haven't visited.

app/src/main/res/layout/poi.xml
```diff
@@ -0,0 +1,47 @@
...
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textSize="15sp" />
+
+            <TextView
+                android:id="@+id/visited_date"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textSize="15sp"/>
+
```
aman3014, line 38: same comment about the visited date

aman3014 (changes requested):

MarcelTorne (approved):

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisRecyclerAdapter.kt
tberkane: Yes i forgot to change this, I converted it to a 'favourited' date. To me it seems like it could be useful to have that.

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisActivity.kt
tberkane: Good catch! I moved the db to the forum's view model. But I did keep this method just for testing purposes.

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisRecyclerAdapter.kt
aman3014: What could the favourite date be useful for?

app/src/main/java/com/github/epfl/meili/profile/favoritepois/FavoritePoisActivity.kt
aman3014: Thanks for changing this! The method seems redundant to me still, why can't we test the db inside the forum's view model?

aman3014 (approved):
Thanks for making the changes :)

PR #208, opened 2021-05-07
Meili Lens

aman3014:
- Landmark detection using Firebase Cloud Vision
- Detection of POI in user's field of view using accelerometer and magnetic field sensors.


MarcelTorne (commented):
Looks awesome @aman3014 ! My only concern is regarding the inheritance of the MapActivityViewModel from PoiMarkerViewModel. I don't think it makes much sense. But I don't have any idea on how to fix it. Maybe having a third view model that will contain PoiMarkerViewModel and LensViewModel? What do you think?
Otherwise the rest is really good! Thanks for the work!!

app/src/main/java/com/github/epfl/meili/map/MapActivityViewModel.kt
```diff
@@ -0,0 +1,147 @@
...
+    private fun registerListener(sensorManager: SensorManager, sensor: Int) {
+        sensorManager.registerListener(
+            MapSensorEventListener(sensor),
+            sensorManager.getDefaultSensor(sensor),
+            SENSOR_DELAY
+        )
+    }
+
+    fun getPOIDist(): LiveData<Pair<PointOfInterest, Int>> = mPOIDist
+    fun getLandmarks(): LiveData<List<FirebaseVisionCloudLandmark>> = mLandMarks
```
tberkane, line 61: Why not put the fields as public instead of getters?

app/src/main/java/com/github/epfl/meili/map/MapActivityViewModel.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    fun handleCameraResponse(uri: Uri) {
```
tberkane, line 63: Missing documentation

app/src/main/java/com/github/epfl/meili/map/MapActivityViewModel.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    private var floatGravity = FloatArray(3)
+    private var floatGeoMagnetic = FloatArray(3)
+
+    private val floatOrientation = FloatArray(3)
+    private val floatRotationMatrix = FloatArray(9)
+
+    private var lastUpdatedAzimuth: Double = 500.0 // impossible azimuth for initialisation
+
+    private val mPOIDist: MutableLiveData<Pair<PointOfInterest, Int>> = MutableLiveData()
```
tberkane, line 43: Nitpick: to be more consistent, use Poi instead of POI

tberkane (changes requested):
Really cool feature, well done!

aman3014:
> Looks awesome @aman3014 ! My only concern is regarding the inheritance of the MapActivityViewModel from PoiMarkerViewModel. I don't think it makes much sense. But I don't have any idea on how to fix it. Maybe having a third view model that will contain PoiMarkerViewModel and LensViewModel? What do you think?

I have fused them into a MapActivityViewModel

app/src/main/java/com/github/epfl/meili/util/LandmarkDetectionService.kt
```diff
@@ -0,0 +1,15 @@
+package com.github.epfl.meili.util
+
+import android.content.Context
+import android.net.Uri
+import com.google.android.gms.tasks.Task
+import com.google.firebase.ml.vision.FirebaseVision
+import com.google.firebase.ml.vision.cloud.landmark.FirebaseVisionCloudLandmark
+import com.google.firebase.ml.vision.common.FirebaseVisionImage
+
+object LandmarkDetectionService {
```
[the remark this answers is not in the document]
aman3014: I don't see a way other than mocking literally everything which doesn't sound too useful. Also I don't see how you could change anything here without having a syntax error :)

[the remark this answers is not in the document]
aman3014: Since the only method of this service is being called from the MapActivityViewModel (which is being tested with a mock landmark detection service).

tberkane (approved):

PR #214, opened 2021-05-10
Bugfix: close keyboard after pressing on any button

tberkane:


aman3014:
Does this only close the keyboard when a button is pressed? But I think when the keyboard is open the buttons are not visible, so I don't think this is what we want.

I think to fix this bug we should close the button when the EditText object loses focus. (I haven't looked into how to do it yet)

tberkane:
> Does this only close the keyboard when a button is pressed? But I think when the keyboard is open the buttons are not visible, so I don't think this is what we want.

Yeah that's true in general but on some screen sizes the buttons are visible. Or with detached keyboard for example.

> I think to fix this bug we should close the button when the EditText object loses focus. (I haven't looked into how to do it yet)

Ah so the 'bug' is that the keyboard should close when user taps an empty spot in the activity? Personally I think that's minor and don't think it's a bug. Lots of apps don't close the keyboard by pressing empty space but you have to use back button of keyboard close button instead, which is what our app does. What do you think?

MarcelTorne (approved):
Thanks for fixing this Thomas! This PR seems to be doing the correct work in my opinion. What I would try to do for future sprints is to specify more the task since otherwise it can be ambiguous what should be done. Finally, I think maybe as a future task we could add that the keyboard is closed whenever you click outside of the keyboard (this is what usually happens in iOS apps at least). Thanks for the work!

aman3014:
I don't really see what this is fixing, since the keyboard is automatically closed when you go to another activity anyways (in the current version on the main).

In the reviews activity, the buttons are not visible when the keyboard is open on my 6.3 inch phone, which I believe is the size of most screens.

aman3014 (changes requested):
Please add this to the forum activity too. And we should remove it from the Profile Activity since this problem doesn't exist there.

(We shouldn't add it to the chat, the keyboard should remain open after sending a message if the user wants to type another message)

tberkane:
> Please add this to the forum activity too.
> 
> (We shouldn't add it to the chat, the keyboard should remain open after sending a message if the user wants to type another message)

That's what I'm doing

aman3014:
I don't see the forum activity being changed.

app/src/main/java/com/github/epfl/meili/profile/ProfileActivity.kt
```diff
@@ -11,6 +11,7 @@ import com.github.epfl.meili.R
 import com.github.epfl.meili.home.Auth
 import com.github.epfl.meili.profile.friends.FriendsListActivity
 import com.github.epfl.meili.util.NavigableActivity
+import com.github.epfl.meili.util.UIUtility
```
aman3014, line 14: unused import

tberkane:
> Thanks for fixing this Thomas! This PR seems to be doing the correct work in my opinion. What I would try to do for future sprints is to specify more the task since otherwise it can be ambiguous what should be done. Finally, I think maybe as a future task we could add that the keyboard is closed whenever you click outside of the keyboard (this is what usually happens in iOS apps at least). Thanks for the work!

Yeah good idea I will add a user story for that

aman3014 (approved):

PR #215, opened 2021-05-10
Fix remarks from code review

tberkane:
Remarks which have not been adressed yet and why:

ProfileActivity.kt
You should not use Deprecated stuff while writing new code.
**-> Looks tricky to fix, don't think it is high priority enough to spend time**

InternetConnectionService.kt
You should not use deprecated stuff when writing new code.
**-> Same**

It is good to have unit tests. You could have more edge cases though, including some tests to trigger errors. This is a good practice to detect bugs.
**-> This is a pretty general remark, if we want to improve this then it's better if each person does it for their tests**

It is not a good practice to use Thread.sleep() in espresso tests. Look at UiController.loopMainThreadUntilIdle() and UiController.loopMainThreadForAtLeast for more information.
**-> The problem is that we use Thread.sleep to wait for a view to load, but those 2 methods require a view to run on.**


aman3014 (commented):
Thanks for addressing these comments!
I have fixed the nitpicks I had.

Question: What is the PoiInfoFragment? Is it a renamed version of something else? I ask because the files changed shows it as a new file being added in this PR.

aman3014:
> PostActivity.kt
Downloading an image from the internet if the first image was not found is not really safe, because one common problem of not being able to download something is not having a connection, thus you will try to download an image from the internet without a connection as well. We recommend downloading it and adding the image as a drawable directly in your app so that it can be displayed locally and won’t require an active internet connection. It would then be a real good default case.

Will be fixed in @Bwananah's PR I believe

tberkane:
> Thanks for addressing these comments!
> I have fixed the nitpicks I had.
> 
> Question: What is the PoiInfoFragment? Is it a renamed version of something else? I ask because the files changed shows it as a new file being added in this PR.

Oh yeah it's a file that we had before but that Marcel removed this week, so I guess I forgot to delete it when merging his branch. I will fix that.

aman3014 (approved):
Looks good to go! Thanks for taking the initiative on this not-so-fun and quite a big PR.

MarcelTorne:
>You should not inject production external service in the tests (see @before in FirebaseAuthenticationServiceTest). You should not test external services (here, as you used mocks, we presume you know that but then, why doing it? As tests are independent, you should not need to first signout as the application would be fresh).
-> Unsure how to fix

So this was almost fixed, there was no need to do sign out since we were already mocking `FirebaseAuth` and `FirebaseAuthenticationService`is our custom class. However, there was a small problem in the timing of setting the mocked `FirebaseAuth`but now its fixed.

MarcelTorne:
> CacheService.kt
setInternetConnectionServicce has a typo and only appear in tests, remove it (unless it would become useful later).
-> Will it be used later?

No, it's fixed now!

aman3014:
> PostActivity.kt
Downloading an image from the internet if the first image was not found is not really safe, because one common problem of not being able to download something is not having a connection, thus you will try to download an image from the internet without a connection as well. We recommend downloading it and adding the image as a drawable directly in your app so that it can be displayed locally and won’t require an active internet connection. It would then be a real good default case.
-> Website which the image comes from was down when I tried it

This is fixed too. I removed the default image, since I'm now realising that it didn't make much sense to have one in the first place.

MarcelTorne (approved):
Thanks for all this tedious work @tberkane ! Made some changes that you can see and we can discuss it if there is any discrepancy. And also added one small comment. Thanks again!

PR #216, opened 2021-05-10
Feature/profile UI

Bwananah:
This PR differentiates between your and other's profiles. It also adds an editable mode to the profile and improves the UI.

By default, you see your profile, but if you putExtra a User using the ProfileActivity's key, you can see their profile.

You can only see people's profiles when connected, and if it is not yours, you can access their comment, post, review and poi history, as well as see their profile name, bio and picture.

On top of that, if it is yours, you can see your friends, edit your profile or sign out.

The big buttons don't do anything for now, but will be used in the future to link to a User's comment, post, review and poi history.


tberkane (approved):
Looks good, good job!

MarcelTorne (commented):
Looks really good to me Ewan! Let me know when you have merged with main and I'll have another look! Thanks for the work!

MarcelTorne (approved):
Awesome job Ewan! Looks really good!

PR #217, opened 2021-05-11
Feature/light dark mode

dyxcooh:
Allows user to choose from dark mode, light mode and follow the system default mode from the profile when logged in.


Bwananah (approved):
Looks good ! Do you think it is possible to improve the test coverage ? I'm not sure if we can test pop ups.

dyxcooh:
> Looks good ! Do you think it is possible to improve the test coverage ? I'm not sure if we can test pop ups.

I tried to click on the alert dialog, but didn't find a correct way to do it

Bwananah:
> 
> 
> > Looks good ! Do you think it is possible to improve the test coverage ? I'm not sure if we can test pop ups.
> 
> I tried to click on the alert dialog, but didn't find a correct way to do it

Ok, no problem ! It doesn't tank our coverage too much so its fine.

aman3014 (commented):
Thanks for this feature! I tried putting the app in dark mode and it works. However, when I restart the app, the app is not in dark mode even though dark mode is selected in the alert dialog. It seems that the app starts in the system default mode no matter what is selected. Would it be possible to add checking the preferences at startup and setting the mode accordingly?

dyxcooh:
> Thanks for this feature! I tried putting the app in dark mode and it works. However, when I restart the app, the app is not in dark mode even though dark mode is selected in the alert dialog. It seems that the app starts in the system default mode no matter what is selected. Would it be possible to add checking the preferences at startup and setting the mode accordingly?

Thank you for pointing this out! I wasn't aware of this behavior because I cannot restart the app on my emulator. I added a checking function which is launched when the profile activity starts and the user is logged in. Or do you mean the check can be done as soon as the app is launched

aman3014:
I was thinking that we could do the check when the app is launched, since the user shouldn't need to be logged in to have the dark mode.

Maybe you could add this check in the map activity?

Also, for now the user can only change the mode if they are logged in if I'm not mistaken, if we add changing the mode to the map activity then this won't be the case anymore.

I would propose to add the change mode button in the [supportActionBar](https://developer.android.com/training/appbar/setting-up) of the MapActivity. But anything should work as long as the user can change the mode without logging in and the mode is set correctly in the main (map) activity.

dyxcooh:
> I was thinking that we could do the check when the app is launched, since the user shouldn't need to be logged in to have the dark mode.
> 
> Maybe you could add this check in the map activity?
> 
> Also, for now the user can only change the mode if they are logged in if I'm not mistaken, if we add changing the mode to the map activity then this won't be the case anymore.
> 
> I would propose to add the change mode button in the [supportActionBar](https://developer.android.com/training/appbar/setting-up) of the MapActivity. But anything should work as long as the user can change the mode without logging in and the mode is set correctly in the main (map) activity.
> 
> We can also ask on the team group for the best place to put change mode button

Yes, it's a good idea to put it in the main page! So now the button is in the action bar next to the camera button. No logged in user can change the mode too

Bwananah (commented):
I made some changes so we have the button in the profile.

aman3014 (approved):
Thanks for this feature @dyxcooh and for your help @Bwananah. I just moved the test to the correct place and ran the linter.

PR #219, opened 2021-05-13
Fix: Now a chat only shows the messages from the current poi and not previously opened ones.

ahmedc3ca:


tberkane (approved):

MarcelTorne (approved):
Thanks for the fix Ahmed!

aman3014:
Thanks for the fix @ahmedc3ca, what was the problem? Just curious :)

PR #237, opened 2021-05-15
Improve UI for Reviews, Forum and Comments to be able to click on authors and go to their profile

MarcelTorne:


Bwananah (changes requested):
Everything looks fine, but when I ran this build on my phone, I couldn't see reviews anymore. 
Everything else seems fine, good job.

ahmedc3ca (commented):
The reviews and comments improvements work perfectly fine. But same as @Bwananah , whenever I open the reviews activity I just get the average of the reviews on the top left but I don't see the individual reviews (I get a storage exception) . Also whenever I open the golf POI (the one we did the demo on) it crashes the app with the exception: _java.lang.IllegalArgumentException: Invalid document reference. Document references must have an even number of segments, but users has 1_ .

MarcelTorne:
> The reviews and comments improvements work perfectly fine. But same as @Bwananah , whenever I open the reviews activity I just get the average of the reviews on the top left but I don't see the individual reviews (I get a storage exception) . Also whenever I open the golf POI (the one we did the demo on) it crashes the app with the exception: _java.lang.IllegalArgumentException: Invalid document reference. Document references must have an even number of segments, but users has 1_ .

Thanks for your comments! I think there are problems with the old reviews (since the arguments are different). I will look more into that and erase them if that's the problem. Because the tests are all passing properly.

MarcelTorne:
> The reviews and comments improvements work perfectly fine. But same as @Bwananah , whenever I open the reviews activity I just get the average of the reviews on the top left but I don't see the individual reviews (I get a storage exception) . Also whenever I open the golf POI (the one we did the demo on) it crashes the app with the exception: _java.lang.IllegalArgumentException: Invalid document reference. Document references must have an even number of segments, but users has 1_ .

After some time trying to find the problem that was causing the problem on the golf poi I realised that this is due to the team members being running different versions of the app and adding new posts. When you add a new post from an old version, there are some attributes missing that then when downloaded from the new version cause the crash in the new version. This is not a big deal, once it is pushed into main, I will delete all posts and then this problem will not appear anymore if we all run the newest version.

Bwananah (approved):
Looks good ! Thanks for your work !

app/src/main/java/com/github/epfl/meili/posts/PostListRecyclerAdapter.kt
```diff
@@ -8,38 +8,42 @@ import android.widget.TextView
 import androidx.recyclerview.widget.RecyclerView
 import com.github.epfl.meili.R
 import com.github.epfl.meili.models.Post
+import com.github.epfl.meili.models.User
+import com.github.epfl.meili.util.ClickListener
 import com.github.epfl.meili.util.MeiliRecyclerAdapter
+import com.github.epfl.meili.util.MeiliWithUserRecyclerViewHolder
+import de.hdodenhof.circleimageview.CircleImageView
```
dyxcooh, line 15: These imports are not used, are you planning to use them later?

app/src/main/java/com/github/epfl/meili/posts/PostActivity.kt
```diff
@@ -90,15 +117,42 @@ class PostActivity : AppCompatActivity() {
...
+        recyclerAdapter.notifyDataSetChanged()
+    }
+
     private fun initRecyclerAdapter() {
-        recyclerAdapter = CommentsRecyclerAdapter(viewModel)
+        recyclerAdapter = CommentsRecyclerAdapter(this)
         val recyclerView: RecyclerView = findViewById(R.id.comments_recycler_view)
-        initRecyclerView(recyclerAdapter, recyclerView,this)
+        initRecyclerView(recyclerAdapter, recyclerView, this)
         ViewCompat.setNestedScrollingEnabled(recyclerView, false);
```
dyxcooh, line 156: Nitpick: redundant semicolon

dyxcooh (approved):
Thanks for this feature, it works perfectly! Just have some nitpicks

ahmedc3ca (approved):

aman3014:
Sorry I clicked on close by mistake

app/src/main/java/com/github/epfl/meili/posts/PostActivity.kt
MarcelTorne: This line wasn't added in this PR but I'll change that. Thanks!

app/src/main/java/com/github/epfl/meili/posts/PostListRecyclerAdapter.kt
MarcelTorne: Removed! Thanks!

PR #238, opened 2021-05-16
Upgrade Firestore Database with the ability to Query

aman3014:
Upgrade FirestoreDatabase with the ability to accept a query, with the the trivial query (always true) being the default one to ensure backwards compatibility.

Changes firestore paths for forum and reviews to make them query-able:
- forum/uid/posts/ -> forum/
- review/uid/reviews/ -> reviews/

The uid of the poi is instead stored inside the post or the review document.


Bwananah (approved):
Looks good ! This will be very useful thanks !

MarcelTorne (approved):
Looks really good! Thanks for the work!

PR #240, opened 2021-05-17
Feed with posts from the nearest points of interests

aman3014:
The feed shows the user the posts from the 10 nearest nearby points of interests, if any.
Refactor the recycler view initialisation to remove duplicate code.

Restructure the posts (previously forum) package given the new general adapter and view model. These should help with future implementations of post lists such as the _My posts_ feature.


Bwananah (approved):
Very clean ! Thanks for also refactoring everything, this will make things easier for a lot of future improvements !

app/src/main/java/com/github/epfl/meili/posts/forum/ForumActivity.kt
```diff
@@ -70,6 +66,7 @@ class ForumActivity : MenuActivity(R.menu.nav_forum_menu), AdapterView.OnItemSel
 
     private lateinit var poi: PointOfInterest
 
+    override fun getActivity(): AppCompatActivity = this
```
MarcelTorne, line 69: One small question, is this function really necessary? Seems redundant to have a function that called on an object that returns the object itself. But maybe I am not understanding the use of it. Could you clarify, please?

app/src/main/java/com/github/epfl/meili/util/PoiServiceViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+    private fun requestPois() {
+        poiService!!.requestPois(
+            lastUserLocation!!,
+            { poiList -> onSuccessPoiReceived(poiList) },
+            { error -> onError(error) })
+    }
+
+    fun onSuccessPoiReceived(poiList: List<PointOfInterest>) {
+        nbCurrentRequests = 0
+    }
```
MarcelTorne, line 37: Was this function originally just setting nbCurrentRequests to 0? If I recall correctly request POIs should accept an onSuccess callback, am I wrong? Not really sure how it is working right now if once you call requestPois there is nothing returned.

MarcelTorne: Okay so I see that you are overriding the function in MapActivity. Wouldn't it be possible to put the observables and the logic behind it here?

app/src/main/java/com/github/epfl/meili/posts/forum/ForumActivity.kt
aman3014: This function is used in the interface `PostListActivity` in functions that need the activity. The other option would be to pass the `this` to the interface's functions when they are called from the activity, but I found this to be the better option. What do you think ?

app/src/main/java/com/github/epfl/meili/util/PoiServiceViewModel.kt
aman3014: So what I was going for here is to have the logic of retrying in the PoiServiceViewModel so that the view models which inherit it don't have to worry about this.

aman3014: What is not ideal in my implementation is that the inheriting view models still need to override the `nbCurrentRequests` variable and set it to 0, but I didn't find a way around this since `PoiServiceViewModel` is an interface.

MarcelTorne: So we finished to discuss this offline, and we decided that it makes more sense to do it as @aman3014 is already doing it right now.

app/src/main/java/com/github/epfl/meili/posts/forum/ForumActivity.kt
MarcelTorne: Awesome! Makes sense! Thanks!

MarcelTorne (approved):
Thanks for your work Aman! Really good job!

PR #241, opened 2021-05-17
Show profile's favorite poi instead of user's

Bwananah:
This PR is a quick bug fix to see profile owner's favourite POIs. Before that, we could only see the app user's.


aman3014 (approved):
Very nice fix, thanks!

dyxcooh (approved):
Good job!

PR #273, opened 2021-05-21
Splash screen & redirection of returning unsigned users directly to the map activity.

aman3014:


Bwananah (approved):
Everything looks good ! Thanks for this feature, it makes the app feel way smoother !

app/src/main/java/com/github/epfl/meili/LogoActivity.kt
```diff
@@ -0,0 +1,40 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import com.github.epfl.meili.home.Auth
+import com.github.epfl.meili.home.FirebaseAuthenticationService
+import com.github.epfl.meili.home.GoogleSignInActivity
+import com.github.epfl.meili.map.MapActivity
+import com.github.epfl.meili.util.UserPreferences
+
+class LogoActivity : AppCompatActivity() {
+    companion object {
+        private const val LOGO_DISPLAY_TIME = 500L // milliseconds
```
ahmedc3ca, line 15: In my opinion, we should either increase this time or remove the app slogan from the starting screen. Since we know what the slogan says we probably read it faster than the average user (especially since revolutionised is such a complicated word).

ahmedc3ca (commented):
Thank you for the addition it makes the experience much smoother! I just have one small comment that I am curious to know your opinion about.

app/src/main/java/com/github/epfl/meili/LogoActivity.kt
aman3014: Sure, I don't really have a strong opinion about this to be honest, let's just ask on the group to see what everyone thinks.

aman3014: The group agrees on 1 second!

ahmedc3ca (approved):
Thank you for the changes.

PR #274, opened 2021-05-21
Improve favorite button

tberkane:
This PR moves the favorite button from the forum activity to the poi info activity. It also makes it a cute heart shape which when toggled adds/deletes the POI from favorites


ahmedc3ca (approved):
Makes sense with the PR #275.  All changes are minor/moving code from one place to the other so no remarks.

Bwananah (approved):
Very clean !

PR #275, opened 2021-05-21
Bugfix: clicking on a poi in favorite pois list now takes you to poi info instead of forum

tberkane:


aman3014:
Thanks for this fix! Since we are improving the UI now, could you also change the button in the profile to reflect that this is the favorite POIs button? For eg. the text could say "Favorite POIs" and the image could be a heart.

tberkane:
> Thanks for this fix! Since we are improving the UI now, could you also change the button in the profile to reflect that this is the favorite POIs button? For eg. the text could say "Favorite POIs" and the image could be a heart.

Good idea, I did that

aman3014 (approved):

ahmedc3ca (commented):
Thank you for the addition, it works as intended! Since now the favourites button in your profile take you to the poi info, maybe we should also move the favourites button in the forum activity to the poi info activity. Does this make sense?

ahmedc3ca (approved):
Thank you for the addition it works as intended!

PR #276, opened 2021-05-21
Restrict posting access to only users who have visited the POI before

MarcelTorne:
In this PR, we are limiting the writing to forum and reviews access to only users that have been in that place before.
Moreover, we are also increasing the distance from which POIs are considered reachable from 50meters to 500 meters.


aman3014 (approved):
Everything looks great, thanks for this feature!

aman3014:
Actually, could you add a simple unit test for the WritingPolicy?

Bwananah (approved):
Well done ! This feature adds a lot to what makes the app special !

PR #277, opened 2021-05-22
Improve camera shutter button placement so that it doesn't block the view in landscape mode

tberkane:


MarcelTorne (commented):
Thanks for this feature @tberkane! I tried to test this but on the emulator I have in my laptop when I put the phone in landscape mode everything appears distortioned, I will leave the review to someone for whom this works well.

ahmedc3ca (approved):
Works as intended 👍

Bwananah (changes requested):
The shutter placement is correct, but now we can't take landscape photos anymore.

tberkane:
> The shutter placement is correct, but now we can't take landscape photos anymore.

Oh yeah good catch. Now it should be good.

Bwananah (commented):
I don't know, I much prefer having the button on the side in landscape mode as it was before, but still be able to take landscape photos. 
Maybe I'm the only one who prefers it that way.
What I imagined was something like Whatsapp does, which I find is the most natural.
Sorry for being picky, I just don't like having the button in the middle in landscape mode.

Bwananah (approved):
Perfect ! Thanks for the changes !

PR #278, opened 2021-05-23
Tabular UI for Poi activities

aman3014:


ahmedc3ca (approved):
Everything looks good and the code looks correct to me! No comment since both HomeActivity and PoiActivity are kind of similar.

Bwananah (approved):
Looks good !

PR #282, opened 2021-05-26
Small fixes to Profile UI

Bwananah:
Improve Profile UI by moving the facebook login button and only having favourite POIs and MyPosts.
I also fixed a bug introduced by the new writable POI feature where favourite POIs couldn't be seen without crashing.


tberkane (approved):
Looks good!

aman3014:
The `poi_history` in the profile XML should be changed to `favorite_pois`, otherwise everything looks good !

aman3014 (approved):
Thanks for these fixes!

PR #283, opened 2021-05-26
Improve Forum and Post UI

Bwananah:
This PR improves the Forum and Post UI by : 
- Placing the upvote and downvote buttons so that they are more reachable and intuitive 
- Moving the add Post button so it is not in an awkward spot
- Removing the default photo from Posts
- Showing photos in the forum preview

It also fixes bugs such as :
- Clearing fields when creating a new post
- UI that didn't scale correctly

/!\ This PR makes changes to the way posts are stored in the database (hasPhoto), so you need a clean forum for it to work /!\


app/src/main/java/com/github/epfl/meili/posts/PostListRecyclerAdapter.kt
```diff
@@ -62,6 +72,16 @@ class PostListRecyclerAdapter(private val viewModel: PostListViewModel, private
...
+
+            if (post.hasPhoto) {
+                FirebaseStorageService.getDownloadUrl(
+                    "images/forum/${post.postId()}",
+                    { uri -> getDownloadUrlCallback(uri) },
+                    { exception ->
+                        Log.e(TAG, "Image not found", exception)
+                    }
+                )
+            }
```
aman3014, line 84: You can use the ImageSetter utility for this

app/src/main/java/com/github/epfl/meili/posts/PostActivity.kt
```diff
@@ -56,10 +56,15 @@ class PostActivity : AppCompatActivity(), UserProfileLinker<Comment>, ClickListe
...
+            FirebaseStorageService.getDownloadUrl(
                 "images/forum/$postId",
-                { uri -> getDownloadUrlCallback(uri) }
-        )
+                { uri -> getDownloadUrlCallback(uri) },
+                { exception ->
+                    Log.e(TAG, "Image not found", exception)
+                }
+            )
+        }
```
aman3014, line 67: Same

app/src/main/java/com/github/epfl/meili/posts/forum/ForumActivity.kt
```diff
@@ -141,17 +141,12 @@ class ForumActivity : PoiActivity(R.layout.activity_forum, R.id.forum_activity),
...
 
-        if (bitmap != null) {
-            executor.execute {
-                compressAndUploadToFirebase(
-                    "images/forum/${post.postId()}",
-                    bitmap!!
-                )
-            }
+        if (post.hasPhoto) {
+            executor.execute { compressAndUploadToFirebase("images/forum/${post.postId()}", bitmap!!) }
```
aman3014, line 149: We should store this path as a const val (since it will be easier to change and appears multiple times)

app/src/main/res/layout/activity_forum.xml
```diff
@@ -1,163 +1,166 @@
...
+
+                <Button
+                    android:id="@+id/cancel_post"
+                    style="@style/Theme.Meili.button"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
                     android:onClick="onClick"
-                    android:contentDescription="@string/use_gallery" />
+                    android:text="@string/cancel"
+                    android:theme="@style/Theme.Meili.button" />
```
aman3014, line 107: Cirrus suggests that the cancel button should always be on the left and the submit on the right.

aman3014 (changes requested):
Thanks for this feature, it makes the forum experience much better!

I just had a couple of refactoring changes to ask before approving.

Also, I think it could be nice to resize the images in the forum. It can be very simply done using Picasso by inserting `resize` between `load` and `into`. What do you think?

MarcelTorne (changes requested):
Thanks for this improvement Ewan! Would it be possible to add some tests to increase code coverage?

Bwananah:
> 
> 
> Thanks for this feature, it makes the forum experience much better!
> 
> I just had a couple of refactoring changes to ask before approving.
> 
> Also, I think it could be nice to resize the images in the forum. It can be very simply done using Picasso by inserting `resize` between `load` and `into`. What do you think?

Maybe in an other PR ? I think it might be difficult to make it look good and see the whole image.

Bwananah:
> 
> 
> Thanks for this improvement Ewan! Would it be possible to add some tests to increase code coverage?

I don't know how to increase test coverage on what I have done, as it is either UI changes or setting images from the database.

aman3014 (approved):
> Maybe in an other PR ? I think it might be difficult to make it look good and see the whole image. Especially because we can have weird image sizes with the cropper.

Okay, makes sense. Thanks for fixing the other issues!

MarcelTorne:
> > Thanks for this improvement Ewan! Would it be possible to add some tests to increase code coverage?
> 
> I don't know how to increase test coverage on what I have done, as it is either UI changes or setting images from the database.

Yeah makes sense, its just that the total coverage is going down by 1% which is weird. Maybe can you try to make some test with a post with a picture? (you can load that picture from the resources, maybe even use the meili app logo to simplify or something) What do you think?

aman3014:
I believe the coverage was only decreasing because the untested image setting code was being duplicated. Since that has been fixed now, it shouldn't be a problem IMO.

MarcelTorne (approved):
Thanks for the work Ewan!

PR #288, opened 2021-05-27
Show poi name of the post in feed and myPosts

dyxcooh:
Display the name of POI for each post when it is viewed in my posts or feed. 
Align 'sort by' text with its spinner


app/src/main/java/com/github/epfl/meili/posts/PostListRecyclerAdapter.kt
```diff
@@ -37,6 +38,7 @@ class PostListRecyclerAdapter(private val viewModel: PostListViewModel, private
             MeiliWithUserRecyclerViewHolder<Post>(itemView, listener) {
         private val title: TextView = itemView.findViewById(R.id.post_title)
         private val postId: TextView = itemView.findViewById(R.id.post_id)
+        private val poiName: TextView = itemView.findViewById(R.id.ppoi_name)
```
aman3014, line 41: typo

app/src/main/java/com/github/epfl/meili/posts/PostListRecyclerAdapter.kt
```diff
@@ -48,8 +50,11 @@ class PostListRecyclerAdapter(private val viewModel: PostListViewModel, private
             super.bind(user, post)
 
             postId.text = post.postId()
+            poiName.text = post.poiName
             title.text = post.title
 
+            if(showPOI){poiName.visibility=View.VISIBLE}
```
aman3014, line 56: autoformatting

aman3014 (changes requested):
Thanks for this feature! I just left a couple of minor remarks.

Also, do you think you could add very simple tests in Feed, Forum, and MyPosts to assert that the poi name is shown or not shown?

dyxcooh:
> Thanks for this feature! I just left a couple of minor remarks.
> 
> Also, do you think you could add very simple tests in Feed, Forum, and MyPosts to assert that the poi name is shown or not shown?

Since the posts are showed as postList in these activities and identified by a single id per list, I cannot access the poi name using an id and do check(matches(isDisplayed())). Do you have an idea about how to test them in mind?

aman3014:
> Since the posts are showed as postList in these activities and identified by a single id per list, I cannot access the poi name using an id and do check(matches(isDisplayed())). Do you have an idea about how to test them in mind?

I think you can use withText instead of withId. Just be careful because it doesn't work if there are multiple matching views. So you'll need to either give different names to pois or make a test with a single poi in the list.

aman3014 (approved):

tberkane (approved):

PR #308, opened 2021-05-28
Fix Meili Lens flicker

aman3014:
Improve POI detection in Meili Lens by increasing FOV and angle tolerance. I also improve the logic for lens poi update (for instance, "catching onto" a new POI is now easier if there is no current POI)

### DIFF
--- a/app/src/main/java/com/github/epfl/meili/map/MapActivityViewModel.kt
+++ b/app/src/main/java/com/github/epfl/meili/map/MapActivityViewModel.kt
@@ -5,6 +5,7 @@ import android.hardware.Sensor
 import android.hardware.SensorEvent
 import android.hardware.SensorEventListener
 import android.hardware.SensorManager
+import android.hardware.SensorManager.SENSOR_DELAY_NORMAL
 import android.location.Location
 import android.net.Uri
 import androidx.core.content.ContextCompat.getSystemService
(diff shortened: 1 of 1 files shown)


MarcelTorne (changes requested):
Thanks for this fix Aman! Would it be possible to add a simple test that will check that everything is updated appropriately when the function is called?

aman3014:
> Thanks for this fix Aman! Would it be possible to add a simple test that will check that everything is updated appropriately when the function is called?

How would you propose testing this? I don't see a way other than making the function public (which is a bad idea since it's not used anywhere else.

MarcelTorne:
> > Thanks for this fix Aman! Would it be possible to add a simple test that will check that everything is updated appropriately when the function is called?
> 
> Do you have any ideas on how we could test this @MarcelTorne ? I don't see a way other than making the function public (which is a bad idea since it's not used anywhere else).

We can create an inheriting lass inside the same test file that can have a public function which will call this private function. It's not the best way but it's the only way I can think about. What do you think?

MarcelTorne (approved):
Discussed offline. Thanks for your work Aman!

tberkane (approved):
A noticeable improvement, thanks! Finding POIs with Meili Lens has never been easier.

PR #314, opened 2021-05-30
Improve tests by making them all have useful assertions

MarcelTorne:
In this PR we are improving the tests codebase so that all of the tests have useful assertions.


aman3014:
> fun createResult(): Result {
        val request: Request = createRequest()
        return Result(request, Result.Code.SUCCESS, createToken(request.permissions), null, null)
    }

I saw that this function in LoginClientCreator is only used in testing, should we move it to the test itself then?

MarcelTorne:
> > fun createResult(): Result {
> > val request: Request = createRequest()
> > return Result(request, Result.Code.SUCCESS, createToken(request.permissions), null, null)
> > }
> 
> I saw that this function in LoginClientCreator is only used in testing, should we move it to the test itself then?

@tberkane and @ahmedc3ca ?

app/src/main/java/com/github/epfl/meili/poi/PoiGoogleRetriever.kt
```diff
@@ -113,7 +113,7 @@ data class GooglePlacesResponse(
     }
 }
 
-//TODO: we could also get some photos from the GooglePlaces API if we want
+//Attention: we could also get some photos from the GooglePlaces API if we want
```
aman3014, line 116: should we just remove this comment?

app/src/main/java/com/github/epfl/meili/util/RotationGestureDetector.kt
```diff
@@ -5,7 +5,7 @@ import kotlin.math.atan2
 
 // Reference used : https://stackoverflow.com/questions/10682019/android-two-finger-rotation
 /** Detects two finger rotation motions */
-class RotationGestureDetector(// Object which can be rotated using two fingers
+open class RotationGestureDetector(// Object which can be rotated using two fingers
```
aman3014, line 8: Why is this class being made open?

app/src/main/java/com/github/epfl/meili/util/RotationGestureDetector.kt
```diff
@@ -31,7 +31,7 @@ class RotationGestureDetector(// Object which can be rotated using two fingers
     }
 
     /** Computes correct indices and positions based on the touch event */
-    fun onTouchEvent(event: MotionEvent): Boolean {
+    open fun onTouchEvent(event: MotionEvent): Boolean {
```
aman3014, line 34: Same question for this function

aman3014:
> Actually this file is already in the test folder. Do you think it is good enough?

Yes! I missed that.

app/src/main/java/com/github/epfl/meili/util/RotationGestureDetector.kt
MarcelTorne: It's a leftover from when I was testing the class, I will remove it.

aman3014 (approved):
Thanks!

Bwananah (approved):
Looks good ! Well done finding a way to test part of the rotation gesture detector ! Thanks !

PR #315, opened 2021-05-30
Sort Comments

Bwananah:
This PR makes it able to sort Comments.

It refactors PostListActivity and PostActivity so they implement an interface called ListSorter, to avoid duplicate code.
I also fixed a bug where posts without images took the images from the previous post, instead of not having any images


tberkane (approved):
:+1:

aman3014:
Looks good, thanks for this feature. I'm just worried that this repeats a lot of code from the post sorting feature. Do you think it would be okay to just have a default newest ordering? In this case no code would be repeated and we would just have to add a sort by timestamp in the listener.

Bwananah:
> 
> 
> Looks good, thanks for this feature. I'm just worried that this repeats a lot of code from the post sorting feature. Do you think it would be okay to just have a default newest ordering? In this case no code would be repeated and we would just have to add a sort by timestamp in the listener.

I made an interface called ListSorter that PostListActivity and PostActivity implement, which reduces duplication. It takes care of the spinner and everything.

aman3014:
> > Looks good, thanks for this feature. I'm just worried that this repeats a lot of code from the post sorting feature. Do you think it would be okay to just have a default newest ordering? In this case no code would be repeated and we would just have to add a sort by timestamp in the listener.
> 
> I made an interface called ListSorter that PostListActivity and PostActivity implement, which reduces duplication. It takes care of the spinner and everything.
> 
> The flags from code climate are the implementations of that function in PostActivity and PostListActivity.

Ah that's awesome then. Sorry, I didn't see your newer changes from yesterday.

aman3014 (approved):
Thanks!

PR #316, opened 2021-05-31
Hide nav bar when typing in POI chat

tberkane:


Bwananah (approved):
Good job ! Thanks !

aman3014:
I'm not sure if I understand why this is a bug. If it is, why is it only being changed for the poi chat and not the friend chat?

tberkane:
> I'm not sure if I understand why this is a bug. If it is, why is it only being changed for the poi chat and not the friend chat?

It's a bug because we don't want to have the nav bar go over the keyboard while typing in the chat. For example, the forum doesn't have this bug because when typing in the fields, the nav bar is hidden. The friend chat is not concerned by this because there is no nav bar in it anyways.

app/src/main/java/com/github/epfl/meili/messages/ChatLogActivity.kt
```diff
@@ -47,13 +52,28 @@ class ChatLogActivity : PoiActivity(R.layout.activity_chat_log, R.id.chat_activi
 
         findViewById<RecyclerView>(R.id.recyclerview_chat_log).adapter = adapter
```
aman3014, line 53: We should store this as an attribute of the class.

app/src/main/java/com/github/epfl/meili/messages/ChatLogActivity.kt
```diff
@@ -47,13 +52,28 @@ class ChatLogActivity : PoiActivity(R.layout.activity_chat_log, R.id.chat_activi
...
+        chatLogView.viewTreeObserver.addOnGlobalLayoutListener {
+            val r = Rect()
+            chatLogView.getWindowVisibleDisplayFrame(r)
+            val heightDiff: Int = chatLogView.rootView
+                .height - (r.bottom - r.top)
+            if (heightDiff > KEYBOARD_THRESHOLD) {
+                navigationBar.visibility = View.GONE
+            } else {
+                navigationBar.visibility = View.VISIBLE
+            }
```
aman3014, line 66: navigationBar.isVisible = heighDiff > KEYBOARD_THRESHOLD

aman3014 (commented):
Okay, makes sense! Just a little change before we merge.

aman3014 (approved):

PR #319, opened 2021-06-04
Several bug fixes

aman3014:


MarcelTorne (approved):

MarcelTorne:
Small comment, there is the review bug that you don't have a different button displayed for modifying your review if you have already added one.

app/src/main/java/com/github/epfl/meili/posts/PostActivity.kt
```diff
@@ -58,10 +58,6 @@ class PostActivity : AppCompatActivity(), ClickListener, ListSorter<Comment> {
         initRecyclerAdapter()
         initLoggedInListener()
         initSorting(findViewById(R.id.sort_spinner), R.array.sort_array_comments)
-
-        findViewById<TextView>(R.id.userName).setOnClickListener {
```
dyxcooh, line 62: Why do you delete this? So we no longer allow user to check the profile from post?

dyxcooh (commented):
Thank you for all these fixes! Looks good for me, I just have one question as commented

app/src/main/java/com/github/epfl/meili/posts/PostActivity.kt
MarcelTorne: No, it was redundant, this is already done in the MeiliRecyclerViewAdapter, but it would be nice if you could double check

dyxcooh: For me it is not working, when I'm on a post and click on the name or photo of the sender, it does not go to the profile

aman3014: Good catch, I have fixed it.

dyxcooh (approved):

PR #320, opened 2021-06-04
Fix name not displaying when logging in with facebook

tberkane:


app/src/main/java/com/github/epfl/meili/auth/SignInActivity.kt
```diff
@@ -6,17 +6,20 @@ import android.view.View
...
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_google_sign_in)
-
-        Auth.setAuthenticationService(FirebaseAuthenticationService())
+        if (AccessToken.getCurrentAccessToken() != null) {
+            Auth.setAuthenticationService(FacebookAuthenticationService())
+        } else
+            Auth.setAuthenticationService(FirebaseAuthenticationService())
```
aman3014: I think this is not required since it's already done in the Logo Activity which always comes before the sign in activity, what do u think?

aman3014 (approved):
Thanks for this fix!

MarcelTorne (approved):
Thanks for fixing this!

PR #321, opened 2021-06-04
fix navigation re-selection bug

aman3014:


MarcelTorne (approved):
Awesome! Thank you Aman!!

tberkane (approved):

